PR #1, opened 2020-06-26
Create

adrianmcli:
Did a bunch of refactoring and added a bunch of warnings.


adrianmcli:
Tasks from feedback:

- [x] make buttons filled instead of outlined
- [x] fix "collateral" spelling at bottom
- [ ] rewrite `If you have an existing position...` to `Maintain 150% collateralization or else you will get liquidated`)
- [ ] add Etherscan links for collateral and tokens: `https://etherscan.io/address/<address>`
- [ ] check for sufficient allowance before enabling create (approval amount must be >= collateral amount)
- [ ] add blurb about shorting by selling minted synths
- [ ] add uniswap link: `https://uniswap.exchange/swap?inputCurrency=<token_address>`

adrianmcli:
New test case:

If it's the user's first position, then check to see that they are minting at least the minSponsorTokens (otherwise dun care).

adrianmcli:
Feature complete, still need to add automated tests + refactor (quite messy right now) but we shouldn't delay merging.

PR #2, opened 2020-07-01
Deposit

chrismaree:
This PR adds the page for deposit.

### DIFF
--- a/containers/Position.ts
+++ b/containers/Position.ts
@@ -3,14 +3,15 @@ import { useState, useEffect } from "react";
 
 import Connection from "./Connection";
 import Contract from "./Contract";
+import { BigNumber } from "ethers";
 
 function usePosition() {
   const { block$, signer, address } = Connection.useContainer();
   const { contract } = Contract.useContainer();
 
-  const [collateral, setCollateral] = useState<string | null>(null);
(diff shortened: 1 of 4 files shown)


adrianmcli:
Build was failing because Typescript thought the component could return `undefined`. It was like this:

```tsx
const MyComponent = () => {
  const x: string = getX()

  if (x == "0") {
    return <div>Foo</div>
  }

  if (x != "0") {
    return <div>Bar</div>
  }
}
```

Although logically speaking, we know that it would always return one of the two, Typescript works by inferring return types _syntactically_ (more [here](https://stackoverflow.com/a/46434101/839793)). That means we always need a final return statement (and remember to use triple-equal signs whenever possible):

```tsx
const MyComponent = () => {
  const x: string = getX()

  // aka the "early-return" or "guard" pattern
  if (x === "0") {
    return <div>Foo</div>
  }

  return <div>Bar</div>
}
```

While there are pros and cons w/ the Early Return pattern, it's used often enough and satisfies the typing requirements here. Here are a couple discussions on the pattern:

- https://medium.com/better-programming/are-early-returns-any-good-eed4b4d03866
- https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

PR #10, opened 2020-07-02
Feat/withdraw

chrismaree:


features/manage-position/Withdraw.tsx
```diff
@@ -0,0 +1,342 @@
...
+    collateral && collateralToWithdraw && tokens
+      ? collateral - parseFloat(collateralToWithdraw) / tokens
+      : startingCR;
+
+  const resultingCRBelowGCR = resultingCR && gcr ? resultingCR < gcr : null;
+
+  const fastWithdrawableCollateral =
+    collateral && tokens && gcr ? (collateral - gcr * tokens).toFixed(2) : null;
+
+  const pendingWithdrawTimeRemaining =
```
chrismaree, line 136: I could not get this to update dynamically. I tried to use a `useEffect` but couldn't get it right.

features/manage-position/Deposit.tsx
```diff
@@ -81,7 +81,7 @@ const Deposit = () => {
         <TextField
           type="number"
           inputProps={{ min: "0" }}
-          label={`Add collateral (${collSymbol})`}
+          label={`Collateral (${collSymbol})`}
```
chrismaree, line 84: Unrelated but cleaned this up

features/manage-position/Withdraw.tsx
```diff
@@ -0,0 +1,342 @@
...
+        <Box py={2}>
+          <Important>
+            IMPORTANT! Please read this carefully or you may lose money.
+          </Important>
+        </Box>
+        <Box pt={2}>
+          <Typography>
+            There are two kinds of withdrawals you can preform:
+          </Typography>
+          <ul style={{ fontSize: 16 }}>
```
chrismaree, line 238: was not sure how to keep the styling the same as the `Typography` section 🤷

features/manage-position/Withdraw.tsx
adrianmcli: `getPosition` ([here](https://github.com/adrianmcli/emp-tools/blob/6de1114160f9f207ce03f02fca3b4c6ba92533ad/containers/Position.ts#L60)) should be called every time there's a new block. Which means it should cause this component to re-render as well.

Is this not the case?

features/manage-position/Withdraw.tsx
adrianmcli: `<Typography>` by default is just a `<p>` tag, this is fine too!

adrianmcli:
This works well enough so I'm going to merge it. We can iron out UX and guard rails later.

PR #11, opened 2020-07-03
Add redeem component

mrice32:
The input validation is a little messy, but everything seems to work fine.

Right now, it's set up to allow the user to type anything they want in the field and to just make the text box red if they type anything that we can't process (I think this is Good Enough™️ for an initial rollout). 

Hopefully, we can figure out how to make the TextField do this validation itself down the road.


adrianmcli:
Yeah I think we should get all 4 features working and "good enough" and then we can spend some time to iron out the guard rails and UX afterwards. Thanks a lot for doing this @mrice32. Btw you can check type errors locally by running `npx tsc`. I should probably add a pre-push hook for that.

PR #12, opened 2020-07-03
Fix minor bugs and add guardrails to prevent reverting txns

mrice32:
Fixes errors in computing the collateralization ratio in Deposit and Withdraw.

Fixes weirdness in displaying a past-expiry time for a withdraw request.

Blocks off other pages when there's a pending withdrawal request (because all other txns will fail).

Adds collateral input validation based on available balance for Create and Deposit.

Adds input validation for cases where the user doesn't have enough synthetic balance to Redeem.


adrianmcli:
Good stuff, thanks @mrice32!

PR #14, opened 2020-07-04
Remove gas limit

mrice32:
This will allow these transactions to estimate their own gas. With a gas limit of 7MM, most of these transactions would ask the user to send $20+ in gas for all of these transactions even though they would really only use a fraction of that.

I'm having trouble testing this in a mainnet fork, but I've done some light testing on mainnet to verify that Ethers correctly estimates gas when this parameter is left out. @adrianmcli, since you have a lot of experience with Ethers, could you verify that this should, indeed, work and give the user a reasonable gas limit?

### DIFF
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -67,9 +67,7 @@ const Create = () => {
       const collateralWei = ethers.utils.parseUnits(collateral);
       const tokensWei = ethers.utils.parseUnits(tokens);
       try {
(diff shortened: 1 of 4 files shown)


adrianmcli:
Gas estimation does not go well with a forked ganache node unfortunately. And yes, all you have to do is leave it out and it should work as normal (i.e. call `eth_estimateGas` on the node and let the node estimate the gas).

Source: https://github.com/ethers-io/ethers.js/issues/194#issuecomment-403336963

PR #16, opened 2020-07-06
Current GCR should have the same precision as user's expected CR

mrice32:
Otherwise, this can create an issue where it appears as if the user's CR is above the GCR, so a fast withdrawal is possible, but the UI doesn't allow you to do it.


adrianmcli:
lgtm

PR #18, opened 2020-07-06
add minSponsorTokens to limit max redeem tokens

### LINKED ISSUE #17: Redeem allows you to send a txn that would take you under the min sponsor size
n/t

adrianmcli:
Fixes #17 

This limits the ability to redeem tokens such that your resulting balance will be less than the `minSponsorTokens` count.

### DIFF
--- a/features/manage-position/Redeem.tsx
+++ b/features/manage-position/Redeem.tsx
@@ -30,9 +30,10 @@ const Redeem = () => {
 
   const tokensToRedeemFloat = isNaN(parseFloat(tokensToRedeem)) ? 0 : parseFloat(tokensToRedeem);
   const tokensAboveMin = borrowedTokens && empState.minSponsorTokens ? borrowedTokens - weiToNum(empState.minSponsorTokens) : 0;
-  const maxRedeem = Math.min((syntheticBalance || 0), (borrowedTokens || 0), (tokensAboveMin ||0));
+  const unwindPosition = borrowedTokens && tokensToRedeemFloat && syntheticBalance && borrowedTokens === tokensToRedeemFloat && borrowedTokens <= syntheticBalance;
+  const maxRedeem = Math.min((syntheticBalance || 0), (borrowedTokens || 0), (tokensAboveMin || 0));
   const isEmpty = tokensToRedeem === "";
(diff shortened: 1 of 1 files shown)


features/manage-position/Redeem.tsx
```diff
@@ -24,7 +29,8 @@ const Redeem = () => {
   const [error, setError] = useState<Error | null>(null);
 
   const tokensToRedeemFloat = isNaN(parseFloat(tokensToRedeem)) ? 0 : parseFloat(tokensToRedeem);
-  const maxRedeem = Math.min((syntheticBalance || 0), (borrowedTokens || 0));
+  const tokensAboveMin = borrowedTokens && empState.minSponsorTokens ? borrowedTokens - weiToNum(empState.minSponsorTokens) : 0;
+  const maxRedeem = Math.min((syntheticBalance || 0), (borrowedTokens || 0), (tokensAboveMin ||0));
```
mrice32, line 33: Does this prevent them from redeeming _all_ of their tokens?

adrianmcli: Yes, I believe so. Should we make an exception for that? Basically the rule should only apply if they are not unwinding their entire position?

adrianmcli: Currently, say `minSponsorTokens` is 1000, and they hold 1200 tokens, then:

- `syntheticBalance` = 1200
- `borrowedTokens` = 1200
- `tokensAboveMin` = 200

They would only be able to redeem 200.

mrice32: I think we should make an exception for that. They should be able to completely unwind their position IMO.

mrice32: So in this case we would want to basically say, you can redeem between 0 and 200 tokens or all 1200 tokens. This makes the logic pretty annoying, though :/

adrianmcli: lol yes this is very annoying.

adrianmcli:
@mrice32 that should do it, it should allow complete unwinding as an "alternative" path.

And yikes, if we're going to support this, we really need to use a proper form validation framework ASAP. 🤮🤮🤮

Issue #7, opened 2020-07-02
Add hyperlinks to etherscan after success tx


chrismaree:
After a successful tx the user is shown their tx hash. This should be a link to etherscan

PR #19, opened 2020-07-06
Add etherscan hyperlink to create transaction hash

nicholaspai:
Init PR adding an etherscan hyperlink to the Create post-transaction message.

If this is a good solution, then will add to other EMP actions

#7

### DIFF
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -10,6 +10,8 @@ import EmpState from "../../containers/EmpState";
 import Totals from "../../containers/Totals";
 import Position from "../../containers/Position";
 
+import { getEtherscanUrl } from "../../utils/getEtherscanUrl";
+
 const Container = styled(Box)`
   max-width: 720px;
 `;
@@ -18,6 +20,11 @@ const Important = styled(Typography)`
   color: red;
   background: black;
   display: inline-block;
+`;
+
+const Link = styled.a`
+  color: white;
+  font-size: 14px;
 `;
 
 const fromWei = ethers.utils.formatUnits;
@@ -240,7 +247,14 @@ const Create = () => {
       {hash && (
         <Box py={2}>
           <Typography>
-            <strong>Tx Hash: </strong> {hash}
+            <strong>Tx Receipt: </strong>
+              <Link
+                href={getEtherscanUrl(hash)}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                {hash}
+              </Link>
           </Typography>
         </Box>
       )}
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -10,7 +10,8 @@ import EmpState from "../../containers/EmpState";
 import Totals from "../../containers/Totals";
 import Position from "../../containers/Position";
 
-import { getEtherscanUrl } from "../../utils/getEtherscanUrl";
+import { useEtherscanUrl } from "../../utils/useEtherscanUrl";
+import { Hash } from "crypto";
 
 const Container = styled(Box)`
   max-width: 720px;
@@ -126,6 +127,8 @@ const Create = () => {
   };
 
   const computedCR = computeCR() || 0;
+
+  const etherscanUrl = useEtherscanUrl(hash);
 
   if (pendingWithdraw === null || pendingWithdraw === "Yes") {
     return (
@@ -246,13 +249,15 @@ const Create = () => {
         <Box py={2}>
           <Typography>
             <strong>Tx Receipt: </strong>
-              <Link
-                href={getEtherscanUrl(hash)}
-                target="_blank"
-                rel="noopener noreferrer"
-              >
-                {hash}
-              </Link>
+              {etherscanUrl ? (
+                <Link
+                  href={etherscanUrl}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  {hash}
+                </Link>
+              ) : hash}
           </Typography>
         </Box>
       )}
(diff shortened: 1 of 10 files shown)


adrianmcli:
There's a type error here (as caught by the CI):

```
13:43:52.161  
ERROR in /vercel/543d99d4/utils/getEtherscanUrl.ts(21,51):
13:43:52.161  
21:51 Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
13:43:52.161  
  Type 'undefined' is not assignable to type 'string'.
13:43:52.161  
    19 | export const getEtherscanUrl = (txnHash: string) => {
13:43:52.161  
    20 |   const { network } = Connection.useContainer();
13:43:52.161  
  > 21 |   const etherscanHost = networkToEtherscanHostMap(network?.name);
13:43:52.161  
       |                                                   ^
13:43:52.161  
    22 | 
13:43:52.161  
    23 |   return `${etherscanHost}/tx/${txnHash}`;
13:43:52.161  
    24 | };
```

btw you can run `npx tsc` to run Typescript which checks the types, but doesn't emit an output (because `noEmit: true` is set in `tsconfig.json`).

features/manage-position/Create.tsx
```diff
@@ -238,7 +245,14 @@ const Create = () => {
       {hash && (
         <Box py={2}>
           <Typography>
-            <strong>Tx Hash: </strong> {hash}
+            <strong>Tx Receipt: </strong>
+              <Link
+                href={getEtherscanUrl(hash)}
```
adrianmcli, line 250: The underlying implementation of `getEtherscanUrl` uses the `Connection` global state container, which is actually a `useContext` hook. Since hooks are not supposed to be called within conditionals, this may cause bugs in the future (if `hash` is null, `getEtherscanUrl` is never called).

I propose that you:

1. Convert `getEtherscanUrl.ts` into a custom hook `useEtherscanUrl.ts`, and;

2. Use it within the component before the returning of JSX.

For example:

```ts
// useEtherscanUrl.ts

import Connection from "../containers/Connection"

const useEtherscanUrl = (hash: string) => {
  const { network } = Connection.useContainer()
  const etherscanUrl =  ... /* insert implementation here */

  return etherscanUrl
}
```

```tsx
// Create.tsx

import useEtherscanUrl from "./useEtherscanUrl"

const Create = () => {
  const etherscanUrl = useEtherscanUrl(hash)
  return (
    <div>{etherscanUrl}</div>
  )
}
```

You'll also have to handle the `null` case, which I agree can be annoying.

nicholaspai: Great suggestion, I implemented this as as suggested.

features/manage-position/Create.tsx
```diff
@@ -238,7 +247,16 @@ const Create = () => {
       {hash && (
         <Box py={2}>
           <Typography>
-            <strong>Tx Hash: </strong> {hash}
+            <strong>Tx Receipt: </strong>
+              {etherscanUrl ? (
```
nicholaspai, line 251: This displays the `etherscanUrl` or the `hash` if the `etherscanUrl` is `undefined` for any reason

utils/useEtherscanUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+        case "goerli":
+            return "https://goerli.etherscan.io";
+        default:
+            return "https://etherscan.io"
+    }
+}
+
+export const useEtherscanUrl = (txnHash: string | null) => {
+    if (!txnHash) { 
+      return undefined;
```
nicholaspai, line 20: This is `undefined` because `Link`'s "href` prop expects `string | undefined`

utils/useEtherscanUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (!txnHash) { 
+      return undefined; 
+    }
+    else {
+        const { network } = Connection.useContainer();
+        const etherscanHost = networkToEtherscanHostMap(network?.name);
+    
+        return `${etherscanHost}/tx/${txnHash}`;  
+    }
+};
```
adrianmcli, line 28:
```suggestion
interface NetworkMap {
  [key: string]: string;
}

const NETWORK_MAP: NetworkMap = {
  kovan: "https://kovan.etherscan.io",
  ropsten: "https://ropsten.etherscan.io",
  rinkeby: "https://rinkeby.etherscan.io",
  goerli: "https://goerli.etherscan.io",
  mainnet: "https://etherscan.io",
};

export const useEtherscanUrl = (txnHash: string | null) => {
  const { network } = Connection.useContainer();
  if (!txnHash || !network) {
    return undefined;
  }

  const baseUrl = NETWORK_MAP[network.name] || NETWORK_MAP["mainnet"];
  return `${baseUrl}/tx/${txnHash}`;
};
```

Reasons for suggested changes:

- Refrain from creating [new functions](https://www.youtube.com/watch?v=Bks59AaHe1c) as they signal complexity and can cause [indirection](https://twitter.com/mpjme/status/814250097651355649). Here the job can be accomplished by a well-typed object literal doing the mapping.

- Hooks (including `Connection.useContainer`) should [never](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) be inside conditionals, so that means the first thing we do is we always call it at the top of our custom hook (which is what this file is).

- Also it's good to always have a return at the end. An `else-return` pattern can sometimes be a code smell as it is frequently [unnecessary](https://eslint.org/docs/rules/no-else-return).

btw, I agree that we should probably fold these guidelines into a linter.

utils/useEtherscanUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+
+module.exports = { useEtherscanUrl };
+
+export {};
```
adrianmcli, line 32:
```suggestion
```

This is not necessary as we already exported the function above with `export const useEtherscanUrl` above.

utils/useEtherscanUrl.ts
nicholaspai: Nice I agree with all of these and especialy like the use of the interface `NetworkMap`

adrianmcli:
clean af, well-done @nicholaspai

PR #24, opened 2020-07-08
Add new error message to Create if user has not selected an EMP

### LINKED ISSUE #23: Show proper messaging on Create when no EMP is selected
1. Connect
2. Go to Manage Position
3. When Create is selected (and no EMP is selected), the following message is displayed:

> You need to cancel or execute your pending withdrawal request before creating additional tokens.

Instead we should probably have a different message if an EMP has not been selected, like:

> Please first select an EMP from the dropdown above

nicholaspai:
Resolves #23 

Open question: Should we add this error message to Deposit/Redeem/Withdraw/Transfer as well? Currently they all display the error message whe `collateral === null` which is something to the effect of "you need a position to do this action".

### DIFF
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -129,6 +129,19 @@ const Create = () => {
 
   const etherscanUrl = useEtherscanUrl(hash);
 
+  // User has not selected an EMP yet. We can detect this by checking if any properties in `empState` are `null`.
+  if (collReq === null) {
+    return (
+      <Container>
+        <Box py={2}>
+          <Typography>
+            <i>Please first select an EMP from the dropdown above.</i>
+          </Typography>
+        </Box>
+      </Container>
+    );
+  }
+
   if (pendingWithdraw === null || pendingWithdraw === "Yes") {
     return (
         <Container>


adrianmcli:
To be honest, I really wanted to maintain [explorability](https://rimble.consensys.design/guides/ux/connect-a-wallet-conditions) by allowing the user to see every single component without having to connect on MetaMask (because they may not have MetaMask installed or they just don't feel like approving this dapp yet).

The downside is that every component needs to have an elegant null state. While I still think this is a worthy goal, it's only practical if we have a dedicated dapp developer w/ dedicated project management in-place. Otherwise, it is probably easier to sacrifice explorability for the sake of maintainability with our skeleton crew and lack of dapp-focus.

Thoughts?

adrianmcli:
Feel free to merge btw.

nicholaspai:
> To be honest, I really wanted to maintain [explorability](https://rimble.consensys.design/guides/ux/connect-a-wallet-conditions) by allowing the user to see every single component without having to connect on MetaMask (because they may not have MetaMask installed or they just don't feel like approving this dapp yet).
> 
> The downside is that every component needs to have an elegant null state, which is easier said than done.
> 
> While I still think this is a worthy goal, it's only practical if we have a dedicated dapp developer w/ dedicated project management in-place. Otherwise, it is probably easier to sacrifice explorability for the sake of maintainability with our skeleton crew and lack of dapp-focus.
> 
> Thoughts?

Agree with maintaining explorability. However, I don't think that this PR affects explorability in any way. Currently if I go to "Manage Position" without connecting Metamask this is what I see:
Screen Shot 2020-07-07 at 22 51 19

adrianmcli:
@nicholaspai my comment on explorability is a bit of a tangent. The deposit/withdraw/redeem functions were implemented without much explorability in mind (as you pointed out). I was just trying to explain why Create still had this null-state: It's because I originally wanted it to be a lot more explorable w/o requiring or asking someone to connect.

PR #26, opened 2020-07-13
New tab "All Positions" displaying information about active sponsor positions

### LINKED ISSUE #25: Display information about sponsor positions
This is valuable for potential liquidators & disputers, and new sponsors who want to have an idea about EMP activity

### LINKED ISSUE #27: Add linting and typescript compilation to Husky precommit hook

nicholaspai:
Resolves #25 

Currently displays:
- Sponsor address
- Collateral locked
- Tokens outstanding
- CR based on off-chain pricefeed (This is a simple call to the Coinbase PRO API for now and is meant to serve as a placeholder)

in a Table:
Screen Shot 2020-07-13 at 15 32 49

Additionally, resolves #27 by adding `lint-staged`, which runs `prettier` on staged files, and typescript compiling to pre-commit Husky hook.


features/all-positions/AllPositions.tsx
```diff
@@ -0,0 +1,86 @@
...
+        );
+    }
+
+    const getEstimatedPrice = () => {
+        if (tokenSymbol?.includes("yCOMP")) {
+            return 187;
+        } else {
+            return 0.0259;
+        }
+    }
```
adrianmcli, line 27:
```suggestion
    const estimatedPrice = tokenSymbol?.includes("yCOMP") ? 187 : 0.0259
```

Of course this also means you can't call it as a function below.

features/all-positions/AllPositions.tsx
```diff
@@ -0,0 +1,86 @@
...
+        createSponsorData('0xD1F55571cbB04139716a9a5076Aa69626B6df009', 350, 170000, 1),
+        createSponsorData('0xd165164cbAb65004Da73C596712687C16b981274', 1, 700, 0),
+        createSponsorData('0x6408c3deE9Aa842db991a1B2fc0ae94D6354724b', 1, 800, 0),
+        createSponsorData('0xD38D3b64504480945F482c03670352CA35679630', 1, 488, 0)
+    ] : [
+        createSponsorData('0x367f62F022E0c8236d664fBA35b594591270Dafb', 1500000, 56250, 2),
+        createSponsorData('0x2e4dE42F0b8ac51D435bd98121Af106388D911Bd', 90000, 4000, 1),
+        createSponsorData('0x74d1bf5ECeEeFc91c61D53c3eDaC9c1173A97853', 500000, 18750, 0),
+        createSponsorData('0xd6D30F186E802c1558b8137bd730f7f4AEC17aE7', 1000, 38, 0),
+        createSponsorData('0x3F603a18Bed7cc5CeEFdC83ff1CE0CF5B3764e2e', 1000, 40, 0)
```
adrianmcli, line 50: what's with these hardcoded values?

features/all-positions/AllPositions.tsx
```diff
@@ -0,0 +1,86 @@
...
+            return 0.0259;
+        }
+    }
+
+    const createSponsorData = (address: string, collateral: number, tokens: number, liqs: number) => {
+        return { 
+            address, 
+            collateral, 
+            tokens, 
+            cRatio: (tokens / (collateral * getEstimatedPrice())),
```
adrianmcli, line 34:
```suggestion
            cRatio: (tokens / (collateral * estimatedPrice)), 
```

features/all-positions/AllPositions.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return (
+        <Box py={4}>
+        <TableContainer component={Paper}>
+        <Table>
+            <TableHead>
+            <TableRow>
+                <TableCell>Active Sponsor</TableCell>
+                <TableCell align="right">Locked Collateral</TableCell>
+                <TableCell align="right">Tokens Outstanding</TableCell>
+                <TableCell align="right">Collateral Ratio (estimated price: {getEstimatedPrice()})</TableCell>
```
adrianmcli, line 63:
```suggestion
                <TableCell align="right">Collateral Ratio (estimated price: {estimatedPrice})</TableCell>
```

features/all-positions/AllPositions.tsx
```diff
@@ -0,0 +1,86 @@
...
+                </TableCell>
+                <TableCell align="right">{row.collateral}</TableCell>
+                <TableCell align="right">{row.tokens}</TableCell>
+                <TableCell align="right">{row.cRatio}</TableCell>
+                <TableCell align="right">{row.liqs}</TableCell>
+                </TableRow>
+            ))}
+            </TableBody>
+        </Table>
+            </TableContainer>
```
adrianmcli, line 81: Try to run the prettier formatter before committing. Perhaps a husky hook for Prettier is necessary.

features/all-positions/AllPositions.tsx
nicholaspai: These are just temporary hardcoded "data" so I can see how the table display looks. I will soon grab these positions dynamicallly haha

features/all-positions/AllPositions.tsx
nicholaspai: This is also a temporary hard coded return val. I will soon make this a dynamic call to an off-chain price feed

features/all-positions/AllPositions.tsx
nicholaspai: Husky hook definitely neccessary lol I can add that

nicholaspai: @adrianmcli what is the prettier formatter command?

package.json
```diff
@@ -36,5 +39,10 @@
     "semi": true,
     "singleQuote": false
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "echo '🛁 Linting changed files with Prettier' && npx pretty-quick --staged && echo '🔏 Compiling Typescript' && npx tsc"
```
adrianmcli, line 44: lol nice touch w/ the emoji

package.json
```diff
@@ -28,6 +28,9 @@
     "babel-plugin-styled-components": "^1.11.0-0",
     "dotenv": "^8.2.0",
     "ganache-core": "^2.10.2",
+    "husky": "^4.2.3",
+    "prettier": "^2.0.5",
+    "pretty-quick": "^2.0.1",
```
adrianmcli, line 33: I wonder if we would want to use `lint-staged` instead in the future (for example, if we want to use ESLint). But it's pretty trivial to switch out.

nicholaspai: I actually think `lint-staged` can run `prettier`

adrianmcli: yeah I meant we might have to replace `pretty-quick` with `lint-staged` if we want to add ESLint later.

nicholaspai: Changed to `lint-staged` which will run `prettier` on staged files

adrianmcli: I think it's fine for now tho

chain.ts
```diff
@@ -5,7 +5,14 @@ import { ethers } from "ethers";
 import { getDai } from "./utils/getDai";
 
 const port = 8545;
+
+// Required environment variables.
 const nodeUrl = process.env.MAINNET_NODE_URL;
+if (!nodeUrl) {
```
nicholaspai, line 11: This might be overkill but it helps for user to know which env vars are required

containers/EmpSponsors.ts
```diff
@@ -0,0 +1,81 @@
...
+
+interface SponsorDictionary {
+  [address: string]: PositionState;
+}
+
+interface EmpDictionary {
+  [address: string]: SponsorDictionary;
+}
+
+const useEmpSponsors = () => {
```
nicholaspai, line 21: This container fetches data for all sponsors that have ever created a position, and currently stores a list of all active sponsors with > 0 locked collateral

containers/PriceFeed.ts
```diff
@@ -0,0 +1,35 @@
...
+
+function usePriceFeed() {
+  const { symbol: tokenSymbol } = Token.useContainer();
+
+  const [latestPrice, setLatestPrice] = useState<BigNumber | null>(null);
+
+  const queryPrice = async () => {
+    if (tokenSymbol) {
+      const query = await getOffchainPrice(
+        tokenSymbol.includes("yCOMP") ? "compusd" : "ethbtc"
```
nicholaspai, line 17: This hard-coded logic is meant to serve as a temporary placeholder for now. The purpose of this container is to house all logic for fetching off-chain prices via API

features/all-positions/AllPositions.tsx
```diff
@@ -0,0 +1,113 @@
...
+  Typography,
+} from "@material-ui/core";
+import { utils, BigNumberish, BigNumber } from "ethers";
+
+import Token from "../../containers/Token";
+import EmpSponsors from "../../containers/EmpSponsors";
+import EmpContract from "../../containers/EmpContract";
+import PriceFeed from "../../containers/PriceFeed";
+
+const AllPositions = () => {
```
nicholaspai, line 20: This file contains all of the code that displays the "All Sponsors" table

nicholaspai:
@mrice32 PTAL at the placeholder price feed logic I have put in place for now.
@adrianmcli and @chrismaree , curious what you think about the husky precommit hook setup

Also, this PR included my `yarn.lock` file. I know we've been back and forth about this but it wasn't on the `gitignore` so I added for now. Happy to remove if that's what our goal is.

utils/getOffchainPrice.ts
```diff
@@ -0,0 +1,30 @@
...
+
+// TODO: This is a really simple mapping of identifier to URL to query to get latest price for an identifier.
+// Future work should blend off-chain prices from different sources similar to how we do it in
+// `protocol/financial-templates-lib/price-feed/MedianizerPriceFeed.js`
+interface PriceFeedIdentifierMap {
+  [identifier: string]: string;
+}
+
+// TODO: Currently using the coinbase API instead of CryptoWatch because CryptoWatch has strict CORS
+// policy that prevents requests coming from localhost domains.
```
mrice32, line 12: This LGTM.

Thoughts on ultimately pulling in `@umaprotocol/financial-templates-lib` down the road to use our common set of price feeds?

utils/getOffchainPrice.ts
```diff
@@ -0,0 +1,30 @@
...
+    const json = await response.json();
+    return utils.parseEther(json[0].price as string);
+  } catch (err) {
+    console.error(`Failed to get price for: ${pricefeedIdentifier}:`, err);
+  }
+};
+
+module.exports = { getOffchainPrice };
+
+export {};
```
mrice32, line 30: nit: just curious -- why is this necessary?

containers/EmpSponsors.ts
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  // get state on setting of contract
+  useEffect(() => {
+    querySponsors();
+  }, [emp]);
+
+  // get state on each block
+  useEffect(() => {
+    if (block$ && emp) {
```
mrice32, line 70: tiny nit: won't querySponsors check for emp != null anyway? So you shouldn't have to worry about it here? This is more just to make sure I'm understanding the logic correctly rather than anything really being wrong with the extra check.

containers/EmpSponsors.ts
```diff
@@ -0,0 +1,81 @@
...
+  lockedCollateral: BigNumber | null;
+}
+
+interface SponsorDictionary {
+  [address: string]: PositionState;
+}
+
+interface EmpDictionary {
+  [address: string]: SponsorDictionary;
+}
```
adrianmcli, line 19: nit (not important at all):

Why not call it `map` instead as it is a "mapping from A to B", `SponsorMap` and `EmpMap` is shorter and simpler as well.

containers/EmpSponsors.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (emp) {
+      const newSponsorFilter = emp.filters.NewSponsor(null);
+      const newSponsorEvents = await emp.queryFilter(newSponsorFilter);
+
+      // Map all active sponsor information to each EMP.
+      if (!activePositions[emp.address]) {
+        activePositions[emp.address] = {};
+      }
+
+      newSponsorEvents.map(async (e: Event) => {
```
adrianmcli, line 38: Should be `.forEach` since you are creating side-effects inside the callback and not making use of the return value.

containers/EmpSponsors.ts
```diff
@@ -0,0 +1,81 @@
...
+  // get position information about every sponsor that has ever created a position.
+  const querySponsors = async () => {
+    if (emp) {
+      const newSponsorFilter = emp.filters.NewSponsor(null);
+      const newSponsorEvents = await emp.queryFilter(newSponsorFilter);
+
+      // Map all active sponsor information to each EMP.
+      if (!activePositions[emp.address]) {
+        activePositions[emp.address] = {};
+      }
```
adrianmcli, line 36: Avoid direct mutation of state variables.

If you need to do things imperatively, make a localized variable for that purpose. e.g.

```ts
let newPositions = { ...activePositions }  // immutably clone object

// perform imperative mutations
if (!newPositions[emp.address]) {
   newPositions[emp.address] = {}
}
```

containers/EmpSponsors.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      newSponsorEvents.map(async (e: Event) => {
+        const sponsorAddress = e.args && e.args.sponsor;
```
adrianmcli, line 39:
```suggestion
        const sponsorAddress = e.args?.sponsor;
```

Nit: This should be equivalent ^

utils/getOffchainPrice.ts
```diff
@@ -0,0 +1,30 @@
+const fetch = require("node-fetch");
```
adrianmcli, line 1: This is not necessary as `fetch` is a native object in the browser API. That's why the name of the package is `node-fetch`, it's only for Node.js environments.

package.json
```diff
@@ -15,6 +15,7 @@
     "ethers": "^5.0.1",
     "jazzicon": "^1.5.0",
     "next": "^9.4.4",
+    "node-fetch": "^2.6.0",
```
adrianmcli, line 18: Don't think this is necessary as mentioned in my other comment.

utils/getOffchainPrice.ts
nicholaspai: Definitely but right now the fact that we have to pass in a Logger object and a web3 object is a bit overkill. I don't know if this repo even has an easy web3 object to access.

utils/getOffchainPrice.ts
nicholaspai: Its not, both the `export{}` and `module.exports{}` are redundant actually---in this and `getDai.ts`

containers/EmpSponsors.ts
nicholaspai: Yeah actually you're right, `querySponsor` won't do anything if `emp === null` so its ok not to check it here.

containers/EmpSponsors.ts
nicholaspai: Will change because Map.length < Dictionary.length, but I've never thought of dicts as different from maps fyi. Have you?

containers/EmpSponsors.ts
nicholaspai: Ah good catch

containers/EmpSponsors.ts
nicholaspai: Good call, I actually clone the `activePositions` mapping later with `updatedPositions` but I should do this earlier before I want to modify the mapping at all.

utils/getOffchainPrice.ts
nicholaspai: Nice

package.json
nicholaspai: Great.

nicholaspai:
@adrianmcli @mrice32 have addressed all your comments PTAL when you have a chance! thanks

containers/EmpSponsors.ts
adrianmcli: oh yeah, they're the same. Subjectively, I just  don't like using the term `dictionary` because it's so many letters and syllables to say (even mentally) lol.

containers/EmpSponsors.ts
```diff
@@ -0,0 +1,81 @@
...
+          const res = await Promise.all([
+            emp.positions(sponsorAddress),
+            emp.getCollateral(sponsorAddress),
+          ]);
+          const positionData = res[0];
+          const collateral = res[1];
+
+          // Active positions have positive locked collateral
+          if (collateral[0].gt(0)) {
+            const updatedPositions = activePositions;
```
adrianmcli: Just a reminder that this doesn't actually clone, it just creates a new reference to the same object. For example:

```js
const a = { foo: "bar" }
const b = a
b.foo = "baz"

console.log(a.foo, b.foo) // baz baz
```

adrianmcli: This is why I did the whole `const a = { ...b }` above to clone.

nicholaspai: Ah I see. Well I did this above now: `let newPositions = { ...activePositions };`

utils/getOffchainPrice.ts
mrice32: This is a good point, and one I hadn't considered. It would be ideal to get all of our libraries to grab default instances of both of these if the user chooses not to provide them.

PR #33, opened 2020-07-16
Use UMA subgraph data for sponsor table

### LINKED ISSUE #32: Use UMA subgraph to query sponsor positions instead of filtering past events
Latest subgraph [explorer](https://thegraph.com/explorer/subgraph/protofire/uma)

GraphQL can be queried via traditional REST API client like `axios` or graphql client like `apollo`:
- example of `axios`: [pools.vision](https://github.com/davis-ramsey/pools-vision/blob/master/src/apis/balancerPools.js)
- example of `apollo`: [uniswap.info](https://github.com/Uniswap/uniswap-info/blob/v2/src/apollo/client.js)

nicholaspai:
Resolves #32 

Stretch TODO: Add a loading graphic while graph query `loading === true`.

### DIFF
--- a/apollo/client.ts
+++ b/apollo/client.ts
@@ -1,0 +1,6 @@
+import { ApolloClient, InMemoryCache } from "@apollo/client";
+
+export const client = new ApolloClient({
+  uri: "https://api.thegraph.com/subgraphs/name/protofire/uma",
+  cache: new InMemoryCache(),
+});
--- a/containers/EmpSponsors.ts
+++ b/containers/EmpSponsors.ts
@@ -33,9 +33,12 @@ interface FinancialContractQuery {
 
 const useEmpSponsors = () => {
   const { contract: emp } = EmpContract.useContainer();
-  const { loading, error, data, networkStatus } = useQuery(ACTIVE_POSITIONS, {
-    pollInterval: 1000,
-    notifyOnNetworkStatusChange: true,
+  // Because apollo caches results of queries, we will poll/refresh this query periodically.
+  // We set the poll interval to a very slow 5 seconds for now since the position states
+  // are not expected to change much.
+  // Source: https://www.apollographql.com/docs/react/data/queries/#polling
+  const { loading, error, data } = useQuery(ACTIVE_POSITIONS, {
+    pollInterval: 5000,
   });
 
   const [activePositions, setActivePositions] = useState<SponsorMap>({});
@@ -69,10 +72,10 @@ const useEmpSponsors = () => {
     }
   };
 
-  // Change state when emp changes or when the graphQL networkStatus changes. This will change every `pollInterval` ms when data is polled anew.
+  // Change state when emp changes or when the graphQL data changes due to polling.
   useEffect(() => {
     querySponsors();
-  }, [emp, networkStatus]);
+  }, [emp, data]);
 
   return { activeSponsors: activePositions };
 };
--- a/containers/EmpSponsors.ts
+++ b/containers/EmpSponsors.ts
@@ -33,7 +33,10 @@ interface FinancialContractQuery {
 
 const useEmpSponsors = () => {
   const { contract: emp } = EmpContract.useContainer();
-  const { loading, error, data } = useQuery(ACTIVE_POSITIONS);
+  const { loading, error, data, networkStatus } = useQuery(ACTIVE_POSITIONS, {
+    pollInterval: 1000,
+    notifyOnNetworkStatusChange: true,
+  });
 
   const [activePositions, setActivePositions] = useState<SponsorMap>({});
 
@@ -66,9 +69,10 @@ const useEmpSponsors = () => {
     }
   };
 
+  // Change state when emp changes or when the graphQL networkStatus changes. This will change every `pollInterval` ms when data is polled anew.
   useEffect(() => {
     querySponsors();
-  }, [emp]);
+  }, [emp, networkStatus]);
 
   return { activeSponsors: activePositions };
 };
--- a/containers/EmpSponsors.ts
+++ b/containers/EmpSponsors.ts
@@ -7,19 +7,28 @@ import EmpContract from "./EmpContract";
 import { useQuery } from "@apollo/client";
 import { ACTIVE_POSITIONS } from "../apollo/queries";
 
+// Interfaces for dApp state storage.
 interface PositionState {
   tokensOutstanding: BigNumberish;
   collateral: BigNumberish;
-  id: string;
+}
+
+interface SponsorPositionState extends PositionState {
+  sponsor: string;
 }
 
 interface SponsorMap {
-  [address: string]: PositionState;
+  [sponsor: string]: SponsorPositionState;
+}
+
+// Interfaces for GraphQl queries.
+interface PositionQuery extends PositionState {
+  sponsor: { id: string };
 }
 
 interface FinancialContractQuery {
   id: string;
-  sponsorPositions: PositionState;
+  sponsorPositions: PositionQuery;
 }
 
 const useEmpSponsors = () => {
@@ -42,13 +51,13 @@ const useEmpSponsors = () => {
             utils.getAddress(contract.id) === emp.address
         );
 
-        empData.sponsorPositions.forEach((position: PositionState) => {
-          const address = utils.getAddress(position.id.split("-")[0]);
+        empData.sponsorPositions.forEach((position: PositionQuery) => {
+          const sponsor = utils.getAddress(position.sponsor.id);
 
-          newPositions[address] = {
+          newPositions[sponsor] = {
             tokensOutstanding: position.tokensOutstanding,
             collateral: position.collateral,
-            id: address,
+            sponsor,
           };
         });
 
--- a/containers/EmpSponsors.ts
+++ b/containers/EmpSponsors.ts
@@ -2,7 +2,6 @@ import { createContainer } from "unstated-next";
 import { useState, useEffect } from "react";
 import { BigNumberish, utils } from "ethers";
 
-import Connection from "./Connection";
 import EmpContract from "./EmpContract";
 
 import { useQuery } from "@apollo/client";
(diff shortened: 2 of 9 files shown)


containers/EmpSponsors.ts
```diff
@@ -18,56 +22,56 @@ interface EmpMap {
...
+          newPositions[emp.address] = {};
+        }
+
+        data.financialContracts
+          .find(
+            (contract: FinancialContractQuery) =>
+              utils.getAddress(contract.id) === emp.address
+          )
+          .sponsorPositions.forEach((position: PositionState) => {
+            const address = utils.getAddress(position.id.split("-")[0]);
```
mrice32, line 58: What is the format of the position `id` field?

containers/EmpSponsors.ts
```diff
@@ -18,56 +22,56 @@ interface EmpMap {
...
 
+interface FinancialContractQuery {
+  id: string;
+  sponsorPositions: PositionState;
+}
+
 const useEmpSponsors = () => {
   const { block$ } = Connection.useContainer();
   const { contract: emp } = EmpContract.useContainer();
+  const { loading, error, data } = useQuery(ACTIVE_POSITIONS);
```
mrice32, line 33: What happens when the state changes? Does this just pull from the cache after the first query and never update again? Or is there a long-running poll that updates `data` every so often?

If data does get updated, does loading ever become `true` again after the first load?

containers/EmpSponsors.ts
```diff
@@ -18,56 +22,56 @@ interface EmpMap {
...
-
-      // Map all active sponsor information to each EMP.
-      if (!newPositions[emp.address]) {
-        newPositions[emp.address] = {};
+      if (error) {
+        console.error(`Apollo client failed to fetch graph data:`, error);
       }
+      if (!loading && data) {
+        // Create a clone of the active-positions mapping that we can update.
+        let newPositions = { ...activePositions };
```
mrice32, line 45: I may be misunderstanding something, but why do we have to start with a clone of the old state? What if a sponsor's position goes from nonzero to 0, will that sponsor continue showing up with their outdated values because they no longer show up in the graph query?

adrianmcli: I'm actually curious about this too. Why not start with a fresh state?

containers/PriceFeed.ts
```diff
@@ -9,7 +9,9 @@ import Token from "./Token";
 function usePriceFeed() {
   const { symbol: tokenSymbol } = Token.useContainer();
 
-  const [latestPrice, setLatestPrice] = useState<BigNumber | null>(null);
+  const [latestPrice, setLatestPrice] = useState<BigNumberish | undefined>(
+    undefined
+  );
```
adrianmcli, line 14: curious why `undefined` is used here instead of `null`

containers/EmpSponsors.ts
```diff
@@ -18,56 +22,56 @@ interface EmpMap {
...
+        if (!newPositions[emp.address]) {
+          newPositions[emp.address] = {};
+        }
+
+        data.financialContracts
+          .find(
+            (contract: FinancialContractQuery) =>
+              utils.getAddress(contract.id) === emp.address
+          )
+          .sponsorPositions.forEach((position: PositionState) => {
```
adrianmcli, line 57: nit: Maybe an intermediate variable could help this looks bit cleaner

```tsx
const empData = data.financialContracts.find(
  (contract: FinancialContractQuery) =>
    utils.getAddress(contract.id) === emp.address
)

empData.forEach(position => {...})
```

features/all-positions/AllPositions.tsx
```diff
@@ -77,37 +75,40 @@ const AllPositions = () => {
...
-                        )
-                      )}
-                    </TableCell>
-                  </TableRow>
-                ))}
+                Object.keys(activeEmpSponsors).map(
+                  (sponsor: string) =>
+                    activeEmpSponsors[sponsor] &&
+                    activeEmpSponsors[sponsor].collateral &&
+                    activeEmpSponsors[sponsor].tokensOutstanding && (
```
adrianmcli, line 88: nit: I notice a lot of usage of `activeEmpSponsors[sponsor]` here, maybe an intermediate variable could help?

```ts
const activeSponsor = activeEmpSponsors[sponsor]
```

containers/EmpSponsors.ts
adrianmcli, line 58: What exactly is `position.id.split("-")[0]`? This line feels a bit magical, it's not exactly easily readable. By that I mean, I have no idea how this line actually gets the `address`.

apollo/client.ts
```diff
@@ -0,0 +1,6 @@
+import { ApolloClient, InMemoryCache } from "@apollo/client";
+
+export const client = new ApolloClient({
+  uri: "https://api.thegraph.com/subgraphs/name/protofire/uma",
+  cache: new InMemoryCache(),
```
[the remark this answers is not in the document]
nicholaspai: Apollo client is graphql client that provides convenient hooks for making graphql queries. Alternatively, we can make REST-like requests and pass in the query in the request body. But these hooks are super useful.

And yes, the cache stores requests. Look [here](https://www.apollographql.com/docs/react/caching/cache-configuration/) for more info.

containers/EmpSponsors.ts
```diff
@@ -18,56 +22,56 @@ interface EmpMap {
...
-      // Create a clone of the active-positions mapping that we can update.
-      let newPositions = { ...activePositions };
-
-      // Map all active sponsor information to each EMP.
-      if (!newPositions[emp.address]) {
-        newPositions[emp.address] = {};
+      if (error) {
+        console.error(`Apollo client failed to fetch graph data:`, error);
       }
+      if (!loading && data) {
```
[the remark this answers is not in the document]
nicholaspai: On mainnet its not empirically faster actually/

containers/EmpSponsors.ts
nicholaspai: The "id" is a string that concatenates the sponsor's address with the financial-contrct's address. We could also grab the sponsor { id } from graph ql instead of just asking for id.

But this made defining the PositionState interface slightly easier (one fewer level)

containers/EmpSponsors.ts
nicholaspai: Actually I'll try this out because I noticed that when switching the EMP's from the dropdown, the table would temporarily show new data and then disappear.

nicholaspai: Made these changes:
- Removed EMPMap, activePositions is now just a SponsorMap since we are only dealing with one EMP at a time.
- Each call to queryPositions starts with a fresh newPositions object.

containers/PriceFeed.ts
nicholaspai: No reason, I'll change to `null`. Curious if you prefer one over the other?

FWIW the `source` has to be undefined because Link's `href` prop expects `string | undefined`

containers/EmpSponsors.ts
nicholaspai: done

features/all-positions/AllPositions.tsx
nicholaspai: Good call

containers/EmpSponsors.ts
nicholaspai: https://github.com/UMAprotocol/emp-tools/pull/33#discussion_r456182918

I'll probably just change the query to explicitly ask for `sponsor { id }` to make it more readable.

containers/PriceFeed.ts
adrianmcli: `null` tends to be a bit more intentional because a lot of things can be `undefined` "naturally" (e.g. referencing a non-existent property in an object), but `null` is usually set by a programmer manually. That being said, it doesn't _really_ matter that much because `null` has other problematic behaviours as well, there really is no right answer here.

A lot of discussion on the web, like this: https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

containers/EmpSponsors.ts
nicholaspai: @adrianmcli @mrice32 I got rid of the `.split("-")[0]` magic and instead just query specifically for the sponsor { id }

containers/EmpSponsors.ts
nicholaspai: > If data does get updated, does loading ever become true again after the first load?

No its only `true` on the first query (i.e. when the cache is empty I'd assume)

> What happens when the state changes? Does this just pull from the cache after the first query and never update again? Or is there a long-running poll that updates data every so often?

I don't think the query will fire again once loaded. But, I think I can set a `poll: ms` option that makes the query refresh periodically. We can also import a `networkStatus` from `useQuery` that will live alert us when polling is done by setting a `notifyOnNetworkStatusChange: true` option. 

I have the `queryPositions()` run now whenever `emp` or `networkStatus` changes. I've set `poll:1000ms`. 

More details [here](https://www.apollographql.com/docs/react/data/queries/#polling)

nicholaspai: Actually: I've decided to set `poll: 5000` since position state shouldn't change at a high frequency.

I also removed the `networkStatus` and the `notifyOnNetWorkStatusChange` vars and instead call `queryPositions()` whenever `emp` or `data` changes.

features/all-positions/AllPositions.tsx
```diff
@@ -68,49 +66,53 @@ const AllPositions = () => {
...
+                  </TableCell>
+                </TableRow>
+              </TableHead>
+              <TableBody>
+                {Object.keys(activeSponsors).map((sponsor: string) => {
+                  const activeSponsor = activeSponsors[sponsor];
+                  return (
+                    activeSponsor &&
+                    activeSponsor.collateral &&
+                    activeSponsor.tokensOutstanding && (
```
adrianmcli, line 89:
```suggestion
                    activeSponsor?.collateral &&
                    activeSponsor?.tokensOutstanding && (
```

I think this is actually sufficient to pass typecheck.

PR #35, opened 2020-07-20
Refine All positions page

chrismaree:
This PR refactors the All positions page to make it a bit more condensed and to add a few extra pieces of info. Previously, the All Positions page looked like this:
image

This PR refines this by shortening sponsor URL, pulling in the symbols for synthetics and debt and changing the wording to make things a bit cleaner. After this PR it looks like this:
image

### DIFF
--- a/features/all-positions/AllPositions.tsx
+++ b/features/all-positions/AllPositions.tsx
@@ -104,7 +104,10 @@ const AllPositions = () => {
                     activeSponsor?.tokensOutstanding && (
                       <TableRow key={sponsor}>
                         <TableCell component="th" scope="row">
-                          <a href={sponsor ? useEtherscanUrl(sponsor) : "N/A"}>
+                          <a
+                            href={sponsor ? useEtherscanUrl(sponsor) : "N/A"}
+                            target="_blank"
+                          >
                             {prettyAddress(sponsor)}
                           </a>
                         </TableCell>
--- a/features/all-positions/AllPositions.tsx
+++ b/features/all-positions/AllPositions.tsx
@@ -50,6 +50,7 @@ const AllPositions = () => {
 
   const prettyBalance = (x: BigNumberish | null) => {
     if (!x) return "N/A";
+    x = Number(x).toFixed(4);
     return utils.commify(x as string);
   };
 
@@ -72,7 +73,8 @@ const AllPositions = () => {
       <Box>
         <Typography>
           <i>
-            Estimated price of {latestPrice ? latestPrice : "N/A"} for{" "}
+            Estimated price of{" "}
+            {latestPrice ? Number(latestPrice).toFixed(4) : "N/A"} for{" "}
             {priceId ? utils.parseBytes32String(priceId) : "N/A"} sourced from{" "}
           </i>
           <Link href={sourceUrl} target="_blank" rel="noopener noreferrer">
--- a/features/all-positions/AllPositions.tsx
+++ b/features/all-positions/AllPositions.tsx
@@ -13,10 +13,14 @@ import {
 import styled from "styled-components";
 import { utils, BigNumberish } from "ethers";
 
+import EmpState from "../../containers/EmpState";
+import Collateral from "../../containers/Collateral";
 import Token from "../../containers/Token";
 import EmpSponsors from "../../containers/EmpSponsors";
 import EmpContract from "../../containers/EmpContract";
 import PriceFeed from "../../containers/PriceFeed";
+
+import { useEtherscanUrl } from "../../utils/useEtherscanUrl";
 
 const Link = styled.a`
   color: white;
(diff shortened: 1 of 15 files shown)


features/all-positions/AllPositions.tsx
```diff
@@ -46,20 +53,28 @@ const AllPositions = () => {
     return utils.commify(x as string);
   };
 
+  const prettyAddress = (x: String | null) => {
+    if (!x) return "N/A";
+    return x.substr(0, 6) + "..." + x.substr(x.length - 6, x.length);
```
nicholaspai, line 58: I like how this makes the address look a lot

features/all-positions/AllPositions.tsx
```diff
@@ -46,20 +53,28 @@ const AllPositions = () => {
...
+  };
+
   const getCollateralRatio = (
     collateral: BigNumberish,
     tokens: BigNumberish
   ) => {
     if (!latestPrice) return null;
     const tokensScaled = Number(tokens) * Number(latestPrice);
-    return Number(collateral) / tokensScaled;
+    return (Number(collateral) / tokensScaled).toFixed(4);
```
nicholaspai, line 67: I think it's worth moving the `.toFixed(4)` into a `prettyRatio()` method because we might want to also use it on the LockedCollateral and TokensOutstanding figures. WDYT?

features/manage-position/Create.tsx
```diff
@@ -10,7 +10,7 @@ import EmpState from "../../containers/EmpState";
 import Totals from "../../containers/Totals";
 import Position from "../../containers/Position";
 
-import { useEtherscanUrl } from "./useEtherscanUrl";
+import { useEtherscanUrl } from "../../utils/useEtherscanUrl";
```
nicholaspai, line 13: I wonder why this diff is showing up I thought it should be merged into `develop` now have you merged with develop since I merged in the WETH PR?

utils/useEtherscanUrl.ts
```diff
@@ -12,12 +12,18 @@ const NETWORK_MAP: NetworkMap = {
...
   const { network } = Connection.useContainer();
-  if (!txnHash || !network) {
+
+  if (!network || !hex || !network.name) {
     return undefined;
   }
-
   const baseUrl = NETWORK_MAP[network.name] || NETWORK_MAP["mainnet"];
-  return `${baseUrl}/tx/${txnHash}`;
+  if (hex.length == 66) {
```
nicholaspai, line 22: 66 and 42 are a bit unreadable to me, can you perhaps define them as constants ADDRESS_HEX_LENGTH and TX_HASH_HEX_LENGTH ?

features/all-positions/AllPositions.tsx
```diff
@@ -46,20 +53,28 @@ const AllPositions = () => {
...
   };
 
   return (
     <Container>
       <Box>
         <Typography>
-          <i>Estimated prices are sourced from the latest trade data on: </i>
+          <i>
+            Estimated price of {latestPrice ? latestPrice : "N/A"} for{" "}
```
nicholaspai, line 75: Can you also round `latestPrice` to 4 decimal places?

features/manage-position/Create.tsx
chrismaree: I made the same changes you did. I'll rebase now and it should remove it.

features/all-positions/AllPositions.tsx
chrismaree: could do that. For now I think it's better to leave the `toFixed` within the given functions as it keeps the logic in the HTML cleaner. I added `toFixed` in the `prettyBalance` and now the table looks like this:
image

Which I think is ideal.

utils/useEtherscanUrl.ts
chrismaree: how about I simply add a comment? having these as extra variables is pretty redundant imho.

features/all-positions/AllPositions.tsx
chrismaree: sure & done :)

features/all-positions/AllPositions.tsx
nicholaspai: >For now I think it's better to leave the toFixed within the given functions as it keeps the logic in the HTML cleaner. I added toFixed in the prettyBalance and now the table looks like this:

Sounds good

utils/useEtherscanUrl.ts
nicholaspai: I just think using variables in functions rather than hard coded values is preferable. I also imagine we will likely use these byte lengths in other future components

adrianmcli: Just throwing in my opinion here, but I think it's fine to have the hardcoded magic numbers since it is pretty clear what it's doing and it's in a small isolated module. It's also something that will never change (i.e. addresses are always going to be 42 in length).

PR #36, opened 2020-07-20
Lint Repo & update package.json

chrismaree:
This PR adds a command to the `package.json` to lint the repo. Along with this, the repo is linted.

### DIFF
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -37,14 +37,14 @@ const Create = () => {
     decimals: collDec,
     allowance: collAllowance,
     setMaxAllowance,
-    balance
+    balance,
   } = Collateral.useContainer();
   const { symbol: tokenSymbol, decimals: tokenDec } = Token.useContainer();
   const { gcr } = Totals.useContainer();
   const {
     collateral: posCollateral,
     tokens: posTokens,
-    pendingWithdraw
+    pendingWithdraw,
   } = Position.useContainer();
 
   const [collateral, setCollateral] = useState<string>("");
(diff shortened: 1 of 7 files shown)


mrice32:
> > LGTM! What about adding a husky pre-commit hook too?
> 
> @nicholaspai had done this already and it works great :)
> I basically just added a command to directly lint the whole repo, rather than just linting staged changes.

ooooh, I see, SGTM! Nice change!

features/manage-position/Create.tsx
```diff
@@ -37,14 +37,14 @@ const Create = () => {
     decimals: collDec,
     allowance: collAllowance,
     setMaxAllowance,
-    balance
+    balance,
```
nicholaspai, line 40: @chrismaree @adrianmcli I've noticed that there are trailing commas like this. Any idea why this is "good practice" lol?

package.json
```diff
@@ -5,7 +5,8 @@
     "dev": "next",
     "build": "next build",
     "start": "next start",
-    "chain": "ts-node -O '{\"module\":\"commonjs\"}' chain.ts"
+    "chain": "ts-node -O '{\"module\":\"commonjs\"}' chain.ts",
+    "lint": "prettier './**/*.{ts,tsx,css,html,md}' --write"
```
nicholaspai, line 9: So you added this because you wanted to be able to run this over the entire repo to capture non-staged files?

chrismaree: yup! I had added some additional functionality but my PR ended up being a lot more than the changes I was making due to the lint diffs :)

PR #38, opened 2020-07-20
Add yToken APY Yield Calculator

### LINKED ISSUE #31: Add y-token yield calculators
Relevant for yCOMP and yUSD

nicholaspai:
Simple feature for now allows user to enter inputs to compute APY.

Resolves #31


features/yield-calculator/YieldCalculator.tsx
```diff
@@ -0,0 +1,120 @@
...
+  margin-bottom: 20px;
+`;
+
+const MS_TO_S = 1000;
+const S_TO_DAYS = 60 * 60 * 24;
+const DAYS_TO_YEAR = 365;
+
+const YieldCalculator = () => {
+  const { empState } = EmpState.useContainer();
+  const [tokenPrice, setTokenPrice] = useState<number | null>(0.9575);
```
mrice32, line 22: nit: where did this magic number come from? Might be preferable to make it a named constant so it's clear that you're using it as the default input price.

features/yield-calculator/YieldCalculator.tsx
```diff
@@ -0,0 +1,120 @@
...
+      const secondsToExpiry = expirationTimestamp.toNumber() - currentTimestamp;
+
+      return Math.round(secondsToExpiry / S_TO_DAYS);
+    } else {
+      return null;
+    }
+  };
+  const [daysToExpiry, setDaysToExpiry] = useState<number>(
+    calculateDaysToExpiry() || 30
+  );
```
mrice32, line 38: High-level comment: the way we're working with daysToExpiry is a little confusing... are we planning on pre-filling it with a calculated value while letting the user change that value? Should we just calculate it for them to make the UI simpler and give them fewer toggles? Not super opinionated here -- just might make this feel a little less technical and more intuitive.

This also may be a WIP, so if so, no worries.

features/yield-calculator/YieldCalculator.tsx
```diff
@@ -0,0 +1,120 @@
...
+  return (
+    <Box pt={4}>
+      <Container>
+        <form noValidate autoComplete="off" onSubmit={(e) => handleClick(e)}>
+          <Typography variant="h5">yUSD Yield Calculator</Typography>
+          <FormInput></FormInput>
+          <FormInput>
+            <TextField
+              type="number"
+              label="Current yUSD Price"
```
mrice32, line 75:
```suggestion
              label="Current yUSD Price (USD)"
```

features/yield-calculator/YieldCalculator.tsx
```diff
@@ -0,0 +1,120 @@
...
+              }
+              variant="outlined"
+              InputLabelProps={{
+                shrink: true,
+              }}
+            />
+          </FormInput>
+          <Button type="submit" variant="contained">
+            Calculate
+          </Button>
```
mrice32, line 113: Other components in EMP tools don't have an explicit button to update information on the page -- they just update as the user types. However, the other components don't exist just to allow the user to calculate something; they exist to submit transactions, so they have a different purpose. Curious to hear your thoughts on an explicit calculate button vs an auto-updating output. Also curious to hear @adrianmcli's thoughts (I don't really have an eye for this sorta stuff).

features/yield-calculator/YieldCalculator.tsx
```diff
@@ -0,0 +1,120 @@
...
+                shrink: true,
+              }}
+            />
+          </FormInput>
+          <FormInput>
+            <TextField
+              type="number"
+              label="Days to Expiry"
+              value={daysToExpiry?.toString() || ""}
+              onChange={(e) => setDaysToExpiry(parseFloat(e.target.value))}
```
mrice32, line 89: This transformation -> float -> input setup causes issues typing some numbers:
- Any number with a `0` after the decimal point is effectively impossible to type.
- Negative numbers (this may be desired).

features/yield-calculator/YieldCalculator.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const calculateYield = () => {
+    if (!tokenPrice || tokenPrice <= 0) {
+      return null;
+    }
+    if (!daysToExpiry || daysToExpiry <= 0) {
+      return null;
+    }
+    const yieldPerUnit =
+      Math.pow(1 / tokenPrice, 1 / (daysToExpiry / DAYS_TO_YEAR)) - 1;
```
mrice32, line 48: nit: might be nice to have a comment briefly explaining the math here -- I'm having trouble intuiting the logic :)

features/yield-calculator/YieldCalculator.tsx
nicholaspai: Its just a random default number that is a placeholder for the live market price of yUSD which we can grab from balancer when the time comes. I'll rename.

features/yield-calculator/YieldCalculator.tsx
nicholaspai: I think we should let the user manually change the daysToExpiry so they can play around with the yield calc.

nicholaspai: I'll rename the 30 magic number to PLACEHOLDER_DAYS_TO_EXPIRY

features/yield-calculator/YieldCalculator.tsx
nicholaspai:
```
// TODO: If the user has not selected an EMP (or has not even connected their web3 provider yet),
// then the EMP's expirationTimestamp will clearly be unavailable. In order to support explorability,
// this is a placeholder `daysToExpiry` that will be the default value for the textfield if no EMP is selected yet.
const PLACEHOLDER_DAYS_TO_EXPIRY = 30;
```

features/yield-calculator/YieldCalculator.tsx
nicholaspai:
```
// TODO: If the user has not selected an EMP (or has not even connected their web3 provider yet),
// then the EMP's expirationTimestamp will clearly be unavailable. In order to support explorability,
// this is a placeholder `daysToExpiry` that will be the default value for the textfield if no EMP is selected yet.
const PLACEHOLDER_DAYS_TO_EXPIRY = 30;
```

features/yield-calculator/YieldCalculator.tsx
nicholaspai:
```
// TODO: When the yUSD market goes live on an exchange, we can replace this
// `PLACEHOLDER_PRICE` with the real-time mid-market price for yUSD.
const PLACEHOLDER_PRICE = 0.9875;
```

features/yield-calculator/YieldCalculator.tsx
nicholaspai: I can auto-update this with a `useEffect()` hook.

features/yield-calculator/YieldCalculator.tsx
nicholaspai: Makes sense. I changed the TextField types for both `daysToExpiry` and `tokenPrice` to string. Please check out the Vercel preview and let me know if it works as desired now.

nicholaspai: >Negative numbers (this may be desired).

Wouldn't negative tokenPrice or daysToExpiry be nonsensical?

features/yield-calculator/YieldCalculator.tsx
nicholaspai: Done

features/yield-calculator/YieldCalculator.tsx
mrice32: It would! That's what I meant by "this may be desired".

features/yield-calculator/YieldCalculator.tsx
```diff
@@ -0,0 +1,133 @@
...
+      const currentTimestamp = Math.round(Date.now() / MS_PER_S);
+      const secondsToExpiry = expirationTimestamp.toNumber() - currentTimestamp;
+
+      return Math.round(secondsToExpiry / S_PER_DAY);
+    } else {
+      return null;
+    }
+  };
+
+  // @dev: We set this state var `daysToExpiry` after declaring `calculateDaysToExpiry()` because we first want to check if its value is
```
mrice32, line 48: nit: is it normal to put `@dev` in front of inline (as opposed to function) comments?

features/yield-calculator/YieldCalculator.tsx
```diff
@@ -0,0 +1,133 @@
...
+              }}
+            />
+          </FormInput>
+          <FormInput>
+            <TextField
+              disabled
+              type="string"
+              label="APY (%)"
+              value={
+                yieldAmount !== null ? `${prettyPercentage(yieldAmount)}` : ""
```
mrice32, line 119: nit: this is the only place you use pretty percentage (I think). It returns "N/A" if the input is `null`. Why not just change it to return `""` if the input is null and avoid handling the null case in both places?

features/yield-calculator/YieldCalculator.tsx
mrice32: SGTM!

features/yield-calculator/YieldCalculator.tsx
mrice32: SGTM! Thanks for the comment!

features/yield-calculator/YieldCalculator.tsx
nicholaspai: No I just do it. I'll remove it

features/yield-calculator/YieldCalculator.tsx
nicholaspai: Great catch. Done

PR #40, opened 2020-07-21
Liquidate and deposit position logic on "All Positions" page

### LINKED ISSUE #43: Add favicon.
Right now the dapp does not have any favicon. Current candidates:
1. 🏗
2. 🛠
3. 🧰
4. 🚧
5. 🤑

chrismaree:
This PR adds a "click to liquidate" and "click to deposit" function to the `All Positions` tab in the dapp. 

This PR was delayed as there is no graph implementation on Kovan. This was required to test out some of the more complex logic. In an effort to get this PR closed, I ended up simply _testing on main net_ by depositing into my own position to validate deposit functions and decoding the liquidation transaction data. I also wanted to have a "click to dispute" function, but this can be left for a separate PR.

Screenshots:

**New table + hover color:**
image

**Deposit tab:**
image

**Liquidation tab:**
image

note that the text at the bottom of this screenshot will update dynamically based on the number of transactions the user needs to perform.

**Deadline dropdown**
image

close #43


apollo/uma/queries.ts
```diff
@@ -7,6 +7,9 @@ export const ACTIVE_POSITIONS = gql`
       positions(first: 1000, where: { collateral_gt: 0 }) {
```
chrismaree, line 7: when the graph is deployed on Kovan we can validate the liquidation structure using:
```
liquidations {
          id
        }
```

features/all-positions/AllPositions.tsx
```diff
@@ -179,53 +216,66 @@ const AllPositions = () => {
...
           </Typography>
         </Box>
         <Box pt={4}>
           <TableContainer component={Paper}>
             <Table>
               <TableHead>
                 <TableRow>
-                  <TableCell>Sponsor</TableCell>
+                  <TableCell>
+                    <strong>Sponsor</strong>
```
chrismaree, line 226: I made the headings bold to make them stand out a bit more.

features/all-positions/AllPositions.tsx
```diff
@@ -40,6 +44,27 @@ const ClickableText = styled.span`
...
+  display: inline-block;
+  text-align: center;
+  &:hover {
+    background-color: white;
+    transition: 0.3s;
+    cursor: pointer;
+  }
+`;
+
+const StyledTableRow = styled(TableRow)`
```
nicholaspai, line 61: This just adds the hover effect right?

features/all-positions/AllPositions.tsx
```diff
@@ -267,6 +322,11 @@ const AllPositions = () => {
             </PageButtons>
           </TableContainer>
         </Box>
+        <PositionActionsDialog
+          isDialogShowing={isDialogShowing}
+          handleClose={toggleDialog}
```
nicholaspai, line 327: This will close the dialog on click away right?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,582 @@
...
+    setDialogTabIndex(newAlignment);
+  };
+
+  const prettyBalance = (x: number) => {
+    const x_string = x.toFixed(4);
+    return utils.commify(x_string);
+  };
+
+  const prettyAddress = (x: String | null) => {
+    if (!x) return "N/A";
```
nicholaspai, line 120: is checking `!x` ALWAYS equivalent to `x !== null`?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,582 @@
...
+    const pendingWithdrawTimeString =
+      pendingWithdrawTimeRemaining && pendingWithdrawTimeRemaining > 0
+        ? Math.max(0, Math.floor(pendingWithdrawTimeRemaining / 3600)) +
+          ":" +
+          Math.max(0, Math.floor((pendingWithdrawTimeRemaining % 3600) / 60)) +
+          ":" +
+          Math.max(0, (pendingWithdrawTimeRemaining % 3600) % 60)
+        : "None";
+
+    const pendingTransferTimeRemaining =
```
nicholaspai, line 143: For now I see no reason to include transfer logic because the rest of the dApp doesn't treat a position as if it has transfer capabilities. Let's just add to another PR?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,582 @@
...
+  };
+
+  const prettyAddress = (x: String | null) => {
+    if (!x) return "N/A";
+    return x.substr(0, 6) + "..." + x.substr(x.length - 6, x.length);
+  };
+  if (
+    props.selectedSponsor &&
+    activeSponsors[props.selectedSponsor] &&
+    latestPrice &&
```
nicholaspai, line 126: SImilar to above comment, prob safer to check explicitly for (!== null)

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,582 @@
...
+        parseFloat(fromWei(collReq)));
+
+    const underCollateralizedPercent =
+      ((Number(latestPrice) - Number(underCollateralizedPrice)) /
+        Number(latestPrice)) *
+      100;
+
+    const collBalanceTooLow = () => {
+      if (dialogTabIndex == "deposit") {
+        return (collBalance || 0) < (Number(collateralToDeposit) || 0);
```
nicholaspai, line 167: I think its more readable if you declare `collBalanceNum = collBalance || 0` at the outset of the file since you do use this logic elsewhere. Same with `collateralToDeposit` and `finalFee` etc.

features/all-positions/AllPositions.tsx
chrismaree: ye

features/all-positions/AllPositions.tsx
chrismaree: yup. if you click away it calls this `toggleDialog` function.

features/all-positions/PositionActionsDialog.tsx
chrismaree: I think so? I'm not 100% sure with typescript. I changed it to be consistent with the rest of the code base for this function.

features/all-positions/PositionActionsDialog.tsx
chrismaree: someone might have done the transfer themselves directly. this modal is to preform liquidations. this text is hidden In most cases if there is no pending transfer anyways.

features/all-positions/PositionActionsDialog.tsx
chrismaree: sure. that's more consistent with how we do it elsewhere. I've changed this, PTAL.

chrismaree:
@nicholaspai I implemented your nits. PTAL.

features/all-positions/PositionActionsDialog.tsx
mrice32: I think it's equivalent only because there is no valid string that's falsy, so null is the only possible value for that type that could be treated as false (but I could be wrong!).

chrismaree:
@nicholaspai I've merged in your latest branch. PTAL.

containers/EmpSponsors.ts
```diff
@@ -28,6 +28,11 @@ interface PositionQuery {
   sponsor: { id: string };
   tokensOutstanding: string;
   collateral: string;
+  pendingWithdraw: string;
```
nicholaspai, line 31: Note that this object only defines the shape of the Query that you are receiving back from graphql. Therefore, since you are only reading the `position.withdrawalRequestTimestamp` and `position.transferPositionRequestTimestamp`, you don't have to actually set `pendingWithdraw` or `pendingTransfer` here.

Additionally, do you use `withdrawRequestAmount`?

nicholaspai, line 31:
```suggestion
```

containers/EmpSponsors.ts
```diff
@@ -28,6 +28,11 @@ interface PositionQuery {
...
+  pendingTransfer: string;
```
nicholaspai, line 32:
```suggestion
```

containers/EmpSponsors.ts
```diff
@@ -28,6 +28,11 @@ interface PositionQuery {
...
+  withdrawalRequestPassTimestamp: string;
+  withdrawalRequestAmount: string;
```
nicholaspai, line 34:
```suggestion
```

features/all-positions/AllSponsors.tsx
```diff
@@ -40,6 +42,27 @@ const PageButtons = styled.div`
   margin-bottom: 2em;
 `;
 
+const MoreInfo = styled.a`
```
nicholaspai, line 45: This dialog styling is something we should eventually pull out into its own module as this config is something we'll want again I'm sure. Outside scope of PR but worth mentioning

features/all-positions/AllSponsors.tsx
```diff
@@ -162,85 +198,119 @@ const AllSponsors = () => {
...
+                setPage(page === maxPage ? page : page + 1);
+              }}
+            >
+              <Arrow faded={page === maxPage ? true : false}>→</Arrow>
+            </div>
+          </PageButtons>
+        </TableContainer>
+        <PositionActionsDialog
+          isDialogShowing={isDialogShowing}
+          handleClose={toggleDialog}
```
nicholaspai, line 310: Do you anticipate adding other logic into `toggleDialog`? Otherwise why not just change this to `() => setDialogShowing(!dialogShowing)` here?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+  Dialog,
+  Button,
+  TextField,
+  Select,
+  InputLabel,
+  FormControl,
+  MenuItem,
+} from "@material-ui/core";
+
+import ToggleButton from "@material-ui/lab/ToggleButton";
```
nicholaspai, line 20: nit: for consistent styling you should change this to 

```
import {
ToggleButton,
ToggleButtonGroup
}  from "@material-ui/lab/ToggleButtonGroup"

File: features/all-positions/PositionActionsDialog.tsx [hunk 15]
```diff
@@ -0,0 +1,583 @@
...
+  const [minCollPerToken, setMinCollPerToken] = useState<string>("");
+  const [maxCollPerToken, setMaxCollPerToken] = useState<string>("");
+  const [maxTokensToLiquidate, setMaxTokensToLiquidate] = useState<string>("");
+  const [deadline, setDeadline] = useState<string>("");
+
+  const [hash, setHash] = useState<string | null>(null);
+  const [success, setSuccess] = useState<boolean | null>(null);
+  const [error, setError] = useState<Error | null>(null);
+
+  const {
```
nicholaspai, line 98: nit: should move this right below where `empState` is declared

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+    priceIdentifier: priceId,
+    collateralRequirement: collReq,
+    currentTime,
+  } = empState;
+
+  const setDialogTab = (
+    event: MouseEvent<HTMLElement>,
+    newAlignment: string | null
+  ) => {
+    setTabIndex(newAlignment);
```
nicholaspai, line 108: OOC: Why do we need to alloow `tabIndex` to possibly be type `null`?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+  const prettyBalance = (x: number) => {
+    const x_string = x.toFixed(4);
+    return utils.commify(x_string);
+  };
+
+  const prettyAddress = (x: String) => {
+    return x.substr(0, 6) + "..." + x.substr(x.length - 6, x.length);
+  };
+  if (
+    props.selectedSponsor &&
```
nicholaspai, line 120: For consistency:

```suggestion
    props.selectedSponsor  !== null &&
    activeSponsors[props.selectedSponsor] !== null &&
    latestPrice !== null &&
    collReq !== null
```

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+    latestPrice !== null &&
+    collReq
+  ) {
+    const collBalanceNum = collBalance || 0;
+    const tokenBalanceNum = tokenBalance || 0;
+    const finalFeeNum = finalFee || 0;
+    const collateralToDepositNum = Number(collateralToDeposit) || 0;
+    const maxTokensToLiquidateNum = Number(maxTokensToLiquidate) || 0;
+
+    const sponsorPosition = activeSponsors[props.selectedSponsor];
```
nicholaspai, line 131: What happens if `props.selectedSponsor === null` or `!sponsorPosition`?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+          ":" +
+          Math.max(0, Math.floor((pendingWithdrawTimeRemaining % 3600) / 60)) +
+          ":" +
+          Math.max(0, (pendingWithdrawTimeRemaining % 3600) % 60)
+        : "None";
+
+    const pendingTransferTimeRemaining =
+      Number(sponsorPosition.TransferTimestamp) - Number(currentTime);
+
+    const pendingTransferTimeString =
```
nicholaspai, line 148:
```suggestion
    const prettyTimeRemaining = (timeRemaining: number) => {
         return timeRemaining > 0
              ? Math.max(0, Math.floor(timeRemaining / 3600)) +
                 ":" +
                 Math.max(0, Math.floor((timeRemaining % 3600) / 60)) +
                 ":" +
                 Math.max(0, (timeRemaining % 3600) % 60)
              : "None";
    }
    const pendingTransferTimeString = prettyTimeRemaining(pendingTransferTimeRemaining)
    const pendingWithdrawTimeString = prettyTimeRemaining(pendingWithdrawTimeRemaining)
```

Might as well not reuse this code block. We use in other modules as well and we can eventually break into its own module

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+      pendingTransferTimeRemaining && pendingTransferTimeRemaining > 0
+        ? Math.max(0, Math.floor(pendingTransferTimeRemaining / 3600)) +
+          ":" +
+          Math.max(0, Math.floor((pendingTransferTimeRemaining % 3600) / 60)) +
+          ":" +
+          Math.max(0, (pendingTransferTimeRemaining % 3600) % 60)
+        : "None";
+
+    const underCollateralizedPrice =
```
nicholaspai, line 157: What if `tokensOutstanding === 0` or `collReq === 0`?

nicholaspai: You might want to default to 0 in that case

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+      Number(sponsorPosition.collateral) /
+      (Number(sponsorPosition.tokensOutstanding) *
+        parseFloat(fromWei(collReq)));
+
+    const underCollateralizedPercent =
```
nicholaspai, line 162: What if `latestPrice === 0`? Again, you might want to default to 0

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+    // Liquidation requires both collateral and synthetic to be approved. This
+    // results in a number of different configurations the user could be in depending
+    // on their interaction with this Dapp. Process the text here to keep the
+    // component simple.
+    const liquidationNeedAllowanceText = () => {
+      if (!needCollateralAllowance() && !needTokenAllowance()) {
+        return null;
+      }
+      if (needCollateralAllowance() && !needTokenAllowance()) {
+        return `You will need to sign two transactions one to approve collateral ${collSymbol} and a second to preform the liquidation.`;
```
nicholaspai, line 206:
```suggestion
        return `You will need to sign two transactions, one to approve collateral ${collSymbol} and a second to perform the liquidation.`;
```

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+      }
+      if (!needCollateralAllowance() && needTokenAllowance()) {
+        return `You will need to sign two transactions one to approve synthetic ${tokenSymbol} and a second to preform the liquidation.`;
```
nicholaspai, line 209:
```suggestion
        return `You will need to sign two transactions, one to approve synthetic ${tokenSymbol} and a second to perform the liquidation.`;
```

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+      }
+      if (needCollateralAllowance() && needTokenAllowance()) {
+        return `You will need to sign three transactions one to approve synthetic ${tokenSymbol}, one to approve collateral ${collSymbol} and a third to preform the liquidation.`;
```
nicholaspai, line 212:
```suggestion
        return `You will need to sign three transactions, one to approve synthetic ${tokenSymbol}, one to approve collateral ${collSymbol} and a third to perform the liquidation.`;
```

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+      }
+    };
+
+    const executeDeposit = async () => {
+      if (collateralToDeposit && !collBalanceTooLow() && emp) {
```
nicholaspai, line 217: Instead of all of these `null` checks, wouldn't it be easier to just not render the entire component unless: 

```
if (emp !== null && tokenAllowance !== null && collateralBalance !== null, ...etc)` 
```

This is running into the same issue that we addressed for all of the existing features in #68.

For example, look at how `Deposit.tsx` does one massive null check at the beginning before rendering the component. This way you don't have to repeatedly check for null variables

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+        setHash(null);
+        setSuccess(null);
+        setError(null);
+        const collateralToDepositWei = toWei(collateralToDeposit);
+
+        try {
+          if (needCollateralAllowance()) await setMaxCollateralAllowance();
+          const tx = await emp.depositTo(props.selectedSponsor, [
```
nicholaspai, line 225: What if `props.selectedSponsor` is null?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+          setError(error);
+        }
+      } else {
+        setError(new Error("Please check that you are connected."));
+      }
+    };
+
+    const executeLiquidation = async () => {
+      if (
+        minCollPerToken &&
```
nicholaspai, line 243: To keep this function logic simpler, I'd recommend disabling the button with `onClick={executeLiquidation` unless all of these strings or other conditionals are defined appropriately. Look at `Deposit.tsx` for an example

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+        <PositionDialog>
+          <Container>
+            <Box>
+              <Box>
+                <h1>Additional Position Actions</h1>
+                <Status>
+                  <Label>Sponsor: </Label>
+                  <a
+                    href={
+                      props.selectedSponsor
```
nicholaspai, line 293: So if you catch `props.selectedSponsor !== undefined` higher up then you won't need these conditionals in the React elements which is easier to maintain going forwards

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+                      {underCollateralizedPrice?.toFixed(4)}. You can still
+                      liquidate this position if you have a different opinion on
+                      the {priceId ? utils.parseBytes32String(priceId) : "N/A"}{" "}
+                      price.
+                      <br></br>
+                      <br></br>
+                    </Typography>
+                    <Important>
+                      Exercise caution! Incorrectly liquidating a position can
+                      loose you money! See the{" "}
```
nicholaspai, line 436:
```suggestion
                      lose you money! See the{" "}
```

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+                      <a
+                        href="https://docs.umaproject.org/synthetic-tokens/explainer#liquidation-and-dispute"
```
nicholaspai, line 438: Should add this to `utils/getDocLinks`

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+                    </Important>
+                    <Box pt={2} pb={2}>
+                      <Grid container spacing={4}>
+                        <Grid item xs={6}>
+                          <TextField
+                            fullWidth
+                            type="number"
+                            inputProps={{ min: "0" }}
+                            label={`Min collateral/token`}
+                            placeholder="1234"
```
nicholaspai, line 454: Do you need this placeholder?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+                        <Grid item xs={6}>
+                          <FormControl fullWidth>
+                            <InputLabel>Deadline</InputLabel>
+                            <Select
+                              value={deadline}
+                              onChange={(
+                                e: React.ChangeEvent<{ value: unknown }>
+                              ) => setDeadline(e.target.value as string)}
+                            >
+                              <MenuItem value={60}>1 minute</MenuItem>
```
nicholaspai, line 502: Should we even give the user the option to select a deadline? Its a bit unclear what the deadline does/how it works, should we not just default it to 30 mins?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,583 @@
...
+                      {minCollPerToken &&
+                      maxCollPerToken &&
+                      maxTokensToLiquidate &&
+                      deadline &&
+                      !collBalanceTooLow() &&
+                      !tokenBalanceTooLow ? (
+                        <Button
+                          fullWidth
+                          variant="contained"
+                          onClick={executeLiquidation}
```
nicholaspai, line 524: This is where I think you can make better use of the `disabled={boolean}` prop and just disable this button unless all of the conditions are set. You decide not to render the button unless:

```
minCollPerToken &&
                      maxCollPerToken &&
                      maxTokensToLiquidate &&
                      deadline &&
                      !collBalanceTooLow() &&
                      !tokenBalanceTooLow ?
```

But I think its better UX to show the user the button in disabled mode. Also, you don't need to catch these conditions a second time in `executeLiquidation`

containers/EmpSponsors.ts
chrismaree: I see, it just shows the structure of the query.

I do use the `withdrawRequestAmount`. it only shows up if the position has a pending withdrawal. The screenshot below shows a position with a withdrawal that has ~2 mins left on it's timer.

image

containers/EmpSponsors.ts
chrismaree: this one I will keep though because it is used & returned from the query.

features/all-positions/AllSponsors.tsx
chrismaree: yup. fair enough. I've refactored this accordingly.

features/all-positions/PositionActionsDialog.tsx
chrismaree: good point. have restricted to be string only.

features/all-positions/PositionActionsDialog.tsx
chrismaree: it cant enter this state due to the if statement that checks `props.selectedSponsor !== null` which wraps all this logic.

features/all-positions/PositionActionsDialog.tsx
chrismaree: ye, good point. Also `sponsorPosition.collateral` could be zero.

features/all-positions/PositionActionsDialog.tsx
chrismaree: I already have the render check set up. I'll expand this accordingly to ensure it covers all variables used

features/all-positions/PositionActionsDialog.tsx
chrismaree: we would not have gotten to this logic as there is the large if check at the top of the rendered component that checks this.

features/all-positions/PositionActionsDialog.tsx
chrismaree: it already checks all this in the button. I've removed the checks here as it's checked there, in the same way that Deposit works.

features/all-positions/PositionActionsDialog.tsx
chrismaree: nope. I've removed them throughout.

features/all-positions/PositionActionsDialog.tsx
chrismaree: yes, I think we should. 30 mins is arbitrary; I'd much rather let the user choose this. They are choosing other complex params in this function, they should also be able to choose this one.

features/all-positions/PositionActionsDialog.tsx
chrismaree: Well, I actually think it's equivalent because in the case that those checks turn out to be false then I render a disabled button.

However, I agree with you that it's cleaner to put this in the `disable` field. rather than having this if/else logic. the only downside is that `disabled` is inherently the negative case so now the whole thing needs to be wrapped in a bracket like:
```
!(
                            minCollPerToken &&
                            maxCollPerToken &&
                            maxTokensToLiquidate &&
                            deadline &&
                            !collBalanceTooLow() &&
                            !tokenBalanceTooLow
                          )
```
which is arguably harder to read anyways.

chrismaree:
@nicholaspai I've implemented your nits. PTAL.

features/all-positions/PositionActionsDialog.tsx
nicholaspai: I just don't want to duplicate the checks and I don't see a problem showing a disabled button

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+      if (tabIndex == "liquidate") {
+        return collBalanceNum < finalFeeNum;
+      }
+    };
+
+    const tokenBalanceTooLow = tokenBalanceNum < maxTokensToLiquidateNum;
+
+    const needCollateralAllowance = () => {
+      if (tabIndex == "deposit") {
+        if (collAllowance === null || collateralToDeposit === null) return true;
```
nicholaspai, line 193: nit: `collAlllowance` cannot be `null` do to the big `if` statement. And you should replace `collateralToDeposit === null` with `collateralToDepositNum === 0`

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+        if (collAllowance === "Infinity") return false;
+        return collAllowance < parseFloat(collateralToDeposit);
```
nicholaspai, line 195:
```suggestion
        return collAllowance < collateralToDepositNum;
```

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+      }
+      if (tabIndex == "liquidate") {
+        if (collAllowance === null || finalFee == null) return true;
```
nicholaspai, line 198: Similar comment as above but with `finalFeeNum`

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+        if (collAllowance === "Infinity") return false;
+        return collAllowance < finalFee;
```
nicholaspai, line 200:
```suggestion
        return collAllowance < finalFeeNum;
```

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+      }
+      return true;
+    };
+    const needTokenAllowance = () => {
+      if (tokenAllowance === null || maxTokensToLiquidate === null) return true;
```
nicholaspai, line 205: Same two comments as above two methods

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+      return tokenAllowance < parseFloat(maxTokensToLiquidate);
+    };
+
+    // Liquidation requires both collateral and synthetic to be approved. This
+    // results in a number of different configurations the user could be in depending
+    // on their interaction with this Dapp. Process the text here to keep the
+    // component simple.
+    const liquidationNeedAllowanceText = () => {
+      if (!needCollateralAllowance() && !needTokenAllowance()) {
+        return null;
```
nicholaspai, line 216: Since this function name implies that you're returning Text, why not return the empty string, "", as the default case?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+        return `You will need to sign three transactions, one to approve synthetic ${tokenSymbol}, one to approve collateral ${collSymbol} and a third to preform the liquidation.`;
+      }
+    };
+
+    const executeDeposit = async () => {
+      if (!collBalanceTooLow()) {
+        setHash(null);
+        setSuccess(null);
+        setError(null);
+        const collateralToDepositWei = toWei(collateralToDeposit);
```
nicholaspai, line 234: Should this be `collateralToDepositNum` instead?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+    };
+
+    const executeLiquidation = async () => {
+      if (!collBalanceTooLow() && !tokenBalanceTooLow) {
+        setHash(null);
+        setSuccess(null);
+        setError(null);
+        const minCollPerTokenWei = toWei(minCollPerToken);
+        const maxCollPerTokenWei = toWei(maxCollPerToken);
+        const maxTokensToLiquidateWei = toWei(maxTokensToLiquidate);
```
nicholaspai, line 261: Should this be `maxTokensToLiquidateNum`

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+                        </Typography>
+                      )}
+                    </Box>
+                  </Box>
+                )}
+                {tabIndex === "liquidate" && (
+                  <Box pt={2}>
+                    <Typography>
+                      <strong>Liquidate this sponsor</strong>
+                      <br></br>For the position to be under collateralize{" "}
```
nicholaspai, line 428:
```suggestion
                      <br></br>For the position to be under collateralized{" "}
```

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+                      {priceId ? utils.parseBytes32String(priceId) : "N/A"}{" "}
+                      would need to{" "}
+                      {underCollateralizedPrice > latestPrice
+                        ? "increase"
+                        : "decrease"}{" "}
+                      by {Math.abs(underCollateralizedPercent | 0)}% from{" "}
```
nicholaspai, line 434: Do you need the `| 0` check, isn't `typeof underCollateralizedPercent === "number"`?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+                      {latestPrice?.toFixed(4)} to{" "}
+                      {underCollateralizedPrice?.toFixed(4)}. You can still
+                      liquidate this position if you have a different opinion on
+                      the {priceId ? utils.parseBytes32String(priceId) : "N/A"}{" "}
```
nicholaspai, line 438: nit: You should add `priceId !== null` to the big null check if statement and set `const priceIdentifier = utils.parseBytes32String(priceId)` right after that check. This is an unneccessarily repeated ternary check

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+                    <Typography>
+                      <strong>Liquidate this sponsor</strong>
+                      <br></br>For the position to be under collateralize{" "}
+                      {priceId ? utils.parseBytes32String(priceId) : "N/A"}{" "}
+                      would need to{" "}
+                      {underCollateralizedPrice > latestPrice
+                        ? "increase"
+                        : "decrease"}{" "}
+                      by {Math.abs(underCollateralizedPercent | 0)}% from{" "}
+                      {latestPrice?.toFixed(4)} to{" "}
```
nicholaspai, line 435: Do you need the ?'s after `latestPrice` and `underCollateralizedPrice` since you already check that they are non-null above?

features/all-positions/PositionActionsDialog.tsx
nicholaspai: Bumping this

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,588 @@
...
+                    </Box>
+
+                    <Box textAlign="center" pt={2} pb={2}>
+                      <Button
+                        fullWidth
+                        variant="contained"
+                        onClick={executeLiquidation}
+                        disabled={
+                          !(
+                            minCollPerToken &&
```
nicholaspai, line 527: Should these be `minCollPerTokenNum > 0 && maxCollPerTokenNum > 0 && etc...`?

features/all-positions/PositionActionsDialog.tsx
chrismaree: went one step further and just removed this fully.

features/all-positions/PositionActionsDialog.tsx
chrismaree: nope. toWei expects a string or bignumber, not an int

features/all-positions/PositionActionsDialog.tsx
chrismaree: same comment as above, cant cast int to big number needs to be string

features/all-positions/PositionActionsDialog.tsx
chrismaree: ye, you right. this is leftover checks from before I set the default to `underCollateralizedPercent` if its empty. have revised.

features/all-positions/PositionActionsDialog.tsx
chrismaree: yup, much cleaner.

features/all-positions/PositionActionsDialog.tsx
chrismaree: not anymore, you right. fixed

features/all-positions/PositionActionsDialog.tsx
chrismaree: na, I dont think so. we dont have the `Num` versions of those for the input types, only for variables that are read in from other state containers. (i.e there is _no_ `minCollPerTokenNum` or `maxCollPerTokenNum`)

chrismaree:
@nicholaspai I've implemented your nits again. we should be good to go now. PTAL.

features/all-positions/AllSponsors.tsx
```diff
@@ -162,85 +194,119 @@ const AllSponsors = () => {
...
+            <div
+              onClick={() => {
+                setPage(page === maxPage ? page : page + 1);
+              }}
+            >
+              <Arrow faded={page === maxPage ? true : false}>→</Arrow>
+            </div>
+          </PageButtons>
+        </TableContainer>
+        <PositionActionsDialog
```
nicholaspai, line 304: You should not render this actions button if `empState.isExpired`

chrismaree:
@nicholaspai I've updated the PR to remove the deposit and liquidate logic if the contract is expired.

You can still click the button to see extra info. Looks like this:

image

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,586 @@
...
+    collSymbol !== null &&
+    collBalance !== null &&
+    collAllowance !== null &&
+    priceId !== null
+  ) {
+    const collBalanceNum = collBalance || 0;
+    const tokenBalanceNum = tokenBalance || 0;
+    const finalFeeNum = finalFee || 0;
+    const collateralToDepositNum = Number(collateralToDeposit) || 0;
+    const maxTokensToLiquidateNum = Number(maxTokensToLiquidate) || 0;
```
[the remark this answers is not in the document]
chrismaree: These two are input props & so are not validated in the render block. They are hooks bound to the input fields.

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,586 @@
...
+    currentTime !== null &&
+    finalFee !== null &&
+    tokenSymbol !== null &&
+    tokenBalance !== null &&
+    tokenAllowance !== null &&
+    collSymbol !== null &&
+    collBalance !== null &&
+    collAllowance !== null &&
+    priceId !== null
+  ) {
```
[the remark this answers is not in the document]
chrismaree: there is no else right now :) 
if _any_ of these are not present then the modal will not open, which is good. There should be nothing displayed until _all_ data has loaded. This is the same pattern we've followed elsewhere so we don't have to validate values are set within the components themselves.

features/all-positions/AllSponsors.tsx
chrismaree: I've chosen to still render the button as it shows additional position info. Rather, I simply hide the liquidation and dispute logic in the dialog if expired.

features/all-positions/AllSponsors.tsx
```diff
@@ -162,85 +194,119 @@ const AllSponsors = () => {
...
-              <TableCell align="right">
-                <SortableTableColumnHeader sortField={SORT_FIELD.CRATIO}>
-                  Collateral Ratio{" "}
-                </SortableTableColumnHeader>
-              </TableCell>
-              <Tooltip
-                title={`This is the price that the identifier (${priceIdUtf8}) must increase to in order for the position be liquidatable`}
-                placement="top"
-              >
+      <div>
```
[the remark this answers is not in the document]
chrismaree: ye, it's a pain. I just made the headings bold and a bit more consistent.

features/all-positions/PositionActionsDialog.tsx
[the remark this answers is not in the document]
nicholaspai: Yeah we've used this pattern before, like [here](https://github.com/UMAprotocol/emp-tools/blob/master/features/manage-position/Create.tsx#L94)

@mrice32

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -0,0 +1,603 @@
...
+  };
+  if (
+    activeSponsors !== null &&
+    props.selectedSponsor !== null &&
+    activeSponsors[props.selectedSponsor] &&
+    activeSponsors[props.selectedSponsor] !== null &&
+    latestPrice !== null &&
+    collReq !== null &&
+    emp !== null &&
+    currentTime !== null &&
```
nicholaspai, line 129: You will want to check if `isExpired !== null` right?

chrismaree: done :)

features/all-positions/PositionActionsDialog.tsx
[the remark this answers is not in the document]
mrice32: Ahh, you're right -- agree on the last two.

However, `collBalance`, `tokenBalance`, and `finalFee` are all checked above, right? So aren't the checks for them redundant?

features/all-positions/PositionActionsDialog.tsx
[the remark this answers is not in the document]
mrice32: Then, can we do:
```
if (!X) {
  return;
}

// Happy path logic
```

features/all-positions/PositionActionsDialog.tsx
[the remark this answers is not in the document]
chrismaree: Ok cool. I've refactored this to only check the input types. PTAL.

PR #44, opened 2020-07-22
Add yUSD and update external price feed + other tidying up tasks

nicholaspai:
Changes:
- Manually add yUSD deployed contract address to list of selectable EMP's
- Replace all displays about "Estimated token price" with "Estimated identifier price" which makes the price feed approximation data point about quanto products (i.e. yUSD collateralized by DAI but using the USD-ETH price feed) more clear what its displaying
- Better mapping of token symbol --> pricefeed URL
- Add `withdrawalLiveness` to EMP state and display in `GeneralInfo` and `Withdraw` features. Also, Withdraw erroneously uses `liquidationLiveness` instead of `withdrawLiveness` to go through request-withdraw flow.
- Dynamically grab etherscan URL's and Doc site URL's. Move `useEtherscanUrl` into its own container `Etherscan` with a `useEtherscanUrl` function.
- From hardcoded EMP address, convert to checksum when setting EMPAddress container


features/contract-state/GeneralInfo.tsx
```diff
@@ -56,7 +59,7 @@ const GeneralInfo = () => {
             target="_blank"
             rel="noopener noreferrer"
           >
-            etherscan
+            Etherscan
```
nicholaspai, line 62: The other links on GeneralInfo are capitalized ("Docs", "Coinbase Pro")

containers/EmpAddress.ts
```diff
@@ -28,7 +28,7 @@ function useEmpAddress() {
   };
 
   return {
-    empAddress: address,
+    empAddress: address ? ethers.utils.getAddress(address) : null,
```
nicholaspai, line 31: Without this, then hardcoding a non-checksummed address in EMPs.ts will cause dApp to crash.

utils/getOffchainPrice.ts
```diff
@@ -1,24 +1,71 @@
...
+  let identifierParams = getIdentifierParamsFromTokenSymbol(symbol);
+  if (!identifierParams) {
+    console.error(
+      `Missing identifier parameters for token with symbol ${symbol}`
+    );
+    return null;
+  }
   try {
-    const response = await fetch(PRICE_FEEDS[pricefeedIdentifier]);
+    const response = await fetch(identifierParams?.pricefeed);
```
mrice32, line 56: Can we lose the `?` here since identifierParams cannot be null at this point? We also omit the `?` below when accessing `.invertedPrice`.

utils/getOffchainPrice.ts
```diff
@@ -1,24 +1,71 @@
...
+  switch (true) {
+    case symbol?.includes("yCOMP"):
+      return IDENTIFIER_PARAMS.compusd;
+    case symbol?.includes("ETHBTC"):
+      return IDENTIFIER_PARAMS.ethbtc;
+    case symbol?.includes("yUSD"):
+      return IDENTIFIER_PARAMS.usdeth;
+    default:
+      return null;
+  }
```
mrice32, line 44: Lol, I had never seen this construction before. Reminds me of the `!!` operator.

features/contract-state/GeneralInfo.tsx
```diff
@@ -76,6 +84,17 @@ const GeneralInfo = () => {
         {priceId ? ethers.utils.parseBytes32String(priceId) : "N/A"}
       </Status>
 
+      <Status>
+        <Label>
+          Estimated Identifier Price (
+          <Link href={sourceUrl} target="_blank" rel="noopener noreferrer">
+            Coinbase Pro
```
mrice32, line 91: If we're pulling the price and the source URL from the PriceFeed container, should we also pull the name from there too?

utils/getOffchainPrice.ts
nicholaspai: Yeah I forgot to link the source : https://stackoverflow.com/questions/4082204/javascript-conditional-switch-statement

Its pretty interesting. The case values have to match the condition in the `switch()` so this evaluates to whichever one is true.

features/contract-state/GeneralInfo.tsx
nicholaspai: Actually we shouldn't really be setting the EMP's identifier in two places since its already being set in the EMPState container which reads from the on-chain contract. I'm going to remove the `identifier` param from the `PriceFeedParams` mapping.

On a separate note, would you prefer to display the identifier's name here in this block? (e.g. "Estimated <IDENTIFIER> Price)

PR #50, opened 2020-07-23
Add dynamic links to exchanges to buy collateral or synthetic tokens

nicholaspai:
Previously we hardcoded the uniswap base URL for an exchange link, this no longer works since yUSD is exchanged on Balancer.

This is potentially way too much work to achieve this functionality, but I created a module `getExchangeLinks` that provides a mapping function from `emp.tokenSymbol ==> getExchangeLink()` where `getExchangeLink()` is a function that takes in some input and returns a URL to the exchange where you can trade for the token symbol. In the future if our tokens trade on other exchanges like Matcha then we can build off of this module to easily add other types of exchanges.

Open to suggestions for a more succinct way of achieving this dynamism.

### DIFF
--- a/utils/getExchangeLinks.ts
+++ b/utils/getExchangeLinks.ts
@@ -33,6 +33,7 @@ export const EXCHANGE_LINK_MAP: ExchangeLinkMap = {
     `https://uniswap.info/token/${tokenAddress}`,
   [EXCHANGES.BALANCER]: (tokenAddress) =>
     `https://balancer.exchange/#/swap/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48/${tokenAddress}`,
+  // This returns a Balancer exchange URL to buy `tokenAddress` and sell USDC, a well understood currency
 };
 
 export const getExchangeTypeFromTokenSymbol = (symbol: string | null) => {
--- a/utils/getExchangeLinks.ts
+++ b/utils/getExchangeLinks.ts
@@ -11,7 +11,7 @@ export enum EXCHANGES {
   BALANCER,
 }
 
-export const EXCHANGE_NAMES: ExchangeNameMap = {
+export const EXCHANGE_NAMES_MAP: ExchangeNameMap = {
   [EXCHANGES.UNISWAP]: "Uniswap",
   [EXCHANGES.BALANCER]: "Balancer",
 };
@@ -30,7 +30,7 @@ export const TOKEN_TO_EXCHANGE_MAP: TokenExchangeMap = {
 
 export const EXCHANGE_LINK_MAP: ExchangeLinkMap = {
   [EXCHANGES.UNISWAP]: (tokenAddress) =>
-    `https://uniswap.info/token/${tokenAddress}`,
+    `https://app.uniswap.org/#/swap?outputCurrency=${tokenAddress}`,
   [EXCHANGES.BALANCER]: (tokenAddress) =>
     `https://balancer.exchange/#/swap/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48/${tokenAddress}`,
   // This returns a Balancer exchange URL to buy `tokenAddress` and sell USDC, a well understood currency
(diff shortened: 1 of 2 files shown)


utils/getExchangeLinks.ts
```diff
@@ -0,0 +1,60 @@
...
+  ethbtc: EXCHANGES.UNISWAP,
+  yusd: EXCHANGES.BALANCER,
+};
+
+export const EXCHANGE_LINK_MAP: ExchangeLinkMap = {
+  [EXCHANGES.UNISWAP]: (tokenAddress) =>
+    `https://uniswap.info/token/${tokenAddress}`,
+  [EXCHANGES.BALANCER]: (tokenAddress) =>
+    `https://balancer.exchange/#/swap/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48/${tokenAddress}`,
+  // This returns a Balancer exchange URL to buy `tokenAddress` and sell USDC, a well understood currency
```
mrice32, line 36: nit: why are we sending them to the balancer exchange, but to uniswap.info? Should we be sending them to the main uniswap site?

utils/getExchangeLinks.ts
```diff
@@ -0,0 +1,60 @@
...
+    case symbol?.includes("ETHBTC"):
+      return TOKEN_TO_EXCHANGE_MAP.ethbtc;
+    case symbol?.includes("yUSD"):
+      return TOKEN_TO_EXCHANGE_MAP.yusd;
+    default:
+      return null;
+  }
+};
+
+export const getExchangeLinkFuncFromTokenSymbol = (symbol: string | null) => {
```
mrice32, line 52: nit: maybe this could be called `getExchange` or `getExchangeInfo` and the `baseurl` return value could be changed to `getTokenUrl` or something that sounds more functiony?

Not super opinionated -- I was just a little confused when I originally saw this. I first assumed that this function returned a single function (rather than an object that had a function as an element), then I assumed that baseurl was just a string (rather than a function).

utils/getExchangeLinks.ts
```diff
@@ -0,0 +1,60 @@
...
+  [EXCHANGES.BALANCER]: "Balancer",
+};
+
+// Returns a function for a given exchange that can be used to retrieve an exchange's specific swap URL for a chosen token.
+interface ExchangeLinkMap {
+  [exchangeEnum: number]: (params: any) => string;
+}
+
+// A mapping of synthetic tokens to the exchange where they can be traded.
+export const TOKEN_TO_EXCHANGE_MAP: TokenExchangeMap = {
```
mrice32, line 25: Do all of these need to be exported? Might be a little simpler for readers/importers if only the functions intended to be used externally are exported? This may also be a ts or js stylistic thing that I'm missing, though.

utils/getExchangeLinks.ts
```diff
@@ -0,0 +1,60 @@
...
+  [EXCHANGES.UNISWAP]: "Uniswap",
+  [EXCHANGES.BALANCER]: "Balancer",
+};
+
+// Returns a function for a given exchange that can be used to retrieve an exchange's specific swap URL for a chosen token.
+interface ExchangeLinkMap {
+  [exchangeEnum: number]: (params: any) => string;
+}
+
+// A mapping of synthetic tokens to the exchange where they can be traded.
```
mrice32, line 24:
```suggestion
// A mapping of synthetic tokens to the exchange where they and their collateral can be traded.
```

utils/getExchangeLinks.ts
nicholaspai: I think we should send them to uniswap.exchange yes. I'll adjust

utils/getExchangeLinks.ts
nicholaspai: I agree we should make `baseurl` sound more function-y. I think `getExchangeInfo` is a lot more clear than the really long name I picked lol

utils/getExchangeLinks.ts
nicholaspai: Nope I meant to remove these. Only `getExchangeInfo` should be exported

utils/getExchangeLinks.ts
nicholaspai: I committed this locally so ignoring the GUI commit.

PR #55, opened 2020-07-23
Fix EMP Management features

### LINKED ISSUE #3: Create does not deal well if you already have a position
In the smart contracts, the `create` function can be called once a use already has a position. This increases their debt and collateral in the same tx. The front end currently does not gracefully deal with the case where the use already has a position. If this is the case they should be told what the resultant CR will be after making the position and show them the delta on their position stats.

The screenshot below shows how this could potentially be appended to the current components:
image

### LINKED ISSUE #45: Creating globally first position is not allowed

### LINKED ISSUE #46: Prevent withdrawals larger than position
image

### LINKED ISSUE #49: Add "liquidation price" to Your Position feature
Show at what price position is liquidatable

### LINKED ISSUE #54: Deposit (and other contract management actions) need to check approval

### LINKED ISSUE #67: Latest build feedback
1. Create refers to "Balance too low" but Deposit refers to "Dai balance too low"
image
image

These should be the same standard. I think bot being the collateral type is good.

2. I can still submit a withdrawal request that would defiantly be liquidated.
I have a position of 412.5 collateral but I can submit a withdrawal request of 400 Dai. The resulting CR of this is 0.0795, which would clearly liquidate me.
image

Ideally, the UI should NOT let me submit any tx that would make my CR < CR requirement.

3. Create shows me a red CR before I've entered any info
image

Here the "CR of newly created tokens" is red. the color here is correct but perhaps only make it red if the value is > 0 & < GCR so it does not start as red by default.

nicholaspai:
- Fixes withdrawal request time remaining counter. Previously was calculating `timeRemaining - withdrawal-liveness`.
- Add liquidation price to sponsor table
- Add liquidation price for potential actions to: Create, Withdraw, Deposit. Redeem doesn't affect liquidation price.
- Adds max "Approve" button to Deposit if user doesn't have enough allowance.
- Removes "Transfer" method selection.
- Adds "Max" button to Redeem so that it can auto-populate the box with 100% of  token balance.
- Disables button for Withdrawals and Creates that would bring CR below CR requirement. Blocks withdrawals for > position collateral size + other withdrawal checks
- Adds liquidation price to YourPosition page.
- Block Redemptions for 0 tokens
- Prevent AllPositions page from crashing if can't find data for EMP on subgraph (this is the case when testing on non-mainnet)
- GCR and CR both default to 0 when tokensOutstanding == 0 instead of null. This allows dApp to display "0" for GCR instead of "N/A" and enables creations on the globally first position (i.e. when GCR should be equal to 0 not `null`)

Resolves #54 
Resolves #49
Resolves #46 
Resolves #3 
Resolves #45 

**!!FYI: If you want to test this easily on Kovan instead of Miannet fork, then I have had good experience (1) replacing the `features/emp-selector/EMPs.ts` list with:**

```
export const EMPs = [
 "0x5FA24Cc5Ea9DB5562113bdf45b81BE7B449D4451", // yUSD Kovan
 "0xFE9fdC3D06189E9bAf3797a7680e96202301e4c3", // yCOMP
 "0xef7C8E8D9F5d5e5BD9e579073A56873B460Cf2E0", // Invalid token with no price feed identifier
];
```

**and (2) changing the WETH contract creation on line ~45 in `containers/WethContract.ts` to:**

```
 const instance = new ethers.Contract(
        "0xd0a1e359811322d97991e03f863a0c30c2cf029c",
        weth.abi,
        signer
      );
```

Update:
- I've completely refactored a large bulk of the feature files to clean up the null checking in the rendering logic. Each feature checks for "valid" variable format (e.g. not null) up front. Within that "if" statement, it then performs the render which makes all of the logic much much cleaner. The now validated data is then either used directly in the `TSX` elements or passed to a `renderComponent` function which returns stateless `TSX` components. There is a very large diff as a result of this refactor. For testing, I'd recommend just playing around as a user on the Vercel preview and trying to break it.

Resolves #67 
Makes a lot of progress toward #68


features/manage-position/Create.tsx
```diff
@@ -240,8 +260,9 @@ const Create = () => {
         )}
         {tokens &&
         collateral &&
-        gcr &&
+        gcr !== null &&
         !needAllowance() &&
+        computedCR !== null &&
```
nicholaspai, line 265: These were preventing creations of the globally first position.

chrismaree:
This PR is awesome! I've gone and played around with it a bunch and have some feedback directly in the code.

As an aside, there is one bug in the redeem tab. This account that I have created 1 yCOMP token but the wallet no longer holds the token. On redeem it shows me this:
image

Which is a bit confusing.

chrismaree:
I like having the additional liquidation price in the table but if you ask me the headings *must not wrap*
image
This means making the columns a bit more efficiently spaced. or, they should wrap consistently. For example `Synthetics (yUSD-SEP20)` should wrap between the `Synthetics` and the `(yUSD-SEP20)` strings. having the break halfway looks a bit janky IMHO.

chrismaree:
The checks on withdrawal are really good now, good work on that. showing the liquidation price is super useful.

features/manage-position/Create.tsx
```diff
@@ -127,11 +133,25 @@ const Create = () => {
...
-  const pricedGCR = gcr && latestPrice ? gcr / Number(latestPrice) : null;
+    latestPrice !== null && latestPrice > 0 && computedCR !== null
+      ? computedCR / Number(latestPrice)
+      : null;
+  const pricedGCR =
+    gcr !== null && latestPrice !== null && latestPrice > 0
+      ? gcr / Number(latestPrice)
+      : null;
+
+  const liquidationPrice =
```
chrismaree, line 147: +1

utils/getLiquidationPrice.ts
```diff
@@ -0,0 +1,20 @@
+export function getLiquidationPrice(
```
chrismaree, line 1: cool to make this a util. There is a BUNCH of other logic we could pull out into common utils like this. I've created an issue to track this: #64

features/manage-position/Withdraw.tsx
```diff
@@ -246,19 +293,19 @@ const Deposit = () => {
           <ul style={{ fontSize: 16 }}>
             <li>
               <strong>"Fast" withdrawal: </strong>Instantly withdraw collateral
```
chrismaree, line 295: this is defs cleaner but do we really want to use acronyms like this?

nicholaspai:
> This PR is awesome! I've gone and played around with it a bunch and have some feedback directly in the code.
> 
> As an aside, there is one bug in the redeem tab. This account that I have created 1 yCOMP token but the wallet no longer holds the token. On redeem it shows me this:
> image
> 
> Which is a bit confusing. I guess it's because my token balance is 0?

Yeah that's correct you have no token balance. We should refine this in a future PR but it functions correctly for now. I don't want to make this PR any more beastly

nicholaspai:
> I like having the additional liquidation price in the table but if you ask me the headings _must not wrap_
> image
> This means making the columns a bit more efficiently spaced. or, they should wrap consistently. For example `Synthetics (yUSD-SEP20)` should wrap between the `Synthetics` and the `(yUSD-SEP20)` strings. having the break halfway looks a bit janky IMHO.
> 
> Also how sensical is the yUSD price to show here? Like what does a liquidation price of `0.0058` mean? I mean I know that `1/0.058=172.41` and so I know that if ETH was to go _below_ 172 this position would get liquidated but this is a lot of steps to jump threw.
> 
> I almost feel that for inverse price feeds like this it doesn't make sense to put this information in the table.

What if I added a tool tip to the column header explaining that the "Liquidation Price is the index (USDETH) price at which the position will get liquidated"

nicholaspai:
> I like having the additional liquidation price in the table but if you ask me the headings _must not wrap_
> image
> This means making the columns a bit more efficiently spaced. or, they should wrap consistently. For example `Synthetics (yUSD-SEP20)` should wrap between the `Synthetics` and the `(yUSD-SEP20)` strings. having the break halfway looks a bit janky IMHO.
> 
> Also how sensical is the yUSD price to show here? Like what does a liquidation price of `0.0058` mean? I mean I know that `1/0.058=172.41` and so I know that if ETH was to go _below_ 172 this position would get liquidated but this is a lot of steps to jump threw.
> 
> I almost feel that for inverse price feeds like this it doesn't make sense to put this information in the table.

I added a `<br>` between "Synthetics/Collateral" and the symbol and this fixes the wrapping as you brough up. However, it still might wrap in the middle of the symbol.

Good display:
Screen Shot 2020-07-24 at 08 37 21

Small display:
Screen Shot 2020-07-24 at 08 37 31

utils/getLiquidationPrice.ts
nicholaspai: Definitely. For example I repeat the calculation `collRequirementFromWei` a lot because the `collReq` in EMPState container is currently stored in Wei form.

features/manage-position/Withdraw.tsx
nicholaspai: I thiink so because on the bottom part we display the CR's with acronyms:
Screen Shot 2020-07-24 at 08 41 35

features/manage-position/Create.tsx
```diff
@@ -109,12 +115,17 @@ const Create = () => {
   };
 
   const computeCR = () => {
+    if (collateral === null || tokens === null) return null;
+    return parseFloat(collateral) / parseFloat(tokens);
+  };
+  const computedCR = computeCR() || 0;
```
nicholaspai, line 121: Need to distinguish between `computedCR` and `resultingCR`.
- `computedCR`: CR of new tokens to create. This must be > GCR.
- `resultingCR`: CR of position including new tokens to create. This can be below GCR but its useful for user to know their new CR.

chrismaree:
> > I like having the additional liquidation price in the table but if you ask me the headings _must not wrap_
> > image
> > This means making the columns a bit more efficiently spaced. or, they should wrap consistently. For example `Synthetics (yUSD-SEP20)` should wrap between the `Synthetics` and the `(yUSD-SEP20)` strings. having the break halfway looks a bit janky IMHO.
> > Also how sensical is the yUSD price to show here? Like what does a liquidation price of `0.0058` mean? I mean I know that `1/0.058=172.41` and so I know that if ETH was to go _below_ 172 this position would get liquidated but this is a lot of steps to jump threw.
> > I almost feel that for inverse price feeds like this it doesn't make sense to put this information in the table.
> 
> I added a `<br>` between "Synthetics/Collateral" and the symbol and this fixes the wrapping as you brough up. However, it still might wrap in the middle of the symbol.
> 
> Good display:
> Screen Shot 2020-07-24 at 08 37 21
> 
> Small display:
> Screen Shot 2020-07-24 at 08 37 31

ye, this looks much better! thanks for making the change.

features/manage-position/Create.tsx
chrismaree: good catch. Bit of a pain given how this contract is implemented but having both cases helps a bunch.

features/manage-position/Create.tsx
```diff
@@ -255,16 +283,27 @@ const Create = () => {
       </Grid>
 
       <Box py={4}>
-        {tokens && collateral && gcr ? (
-          <Typography>
-            Resulting CR:{" "}
-            <span style={{ color: computedCR < gcr ? "red" : "unset" }}>
-              {pricedCR?.toFixed(4)}
+        <Typography>
+          CR of newly created tokens:{" "}
```
mrice32, line 287: Can I ask a super dumb question? What does `{" "}` do? It seems to be this weird placeholder employed throughout jsx and I've never fully understood the purpose.

features/manage-position/ManagePosition.tsx
```diff
@@ -10,7 +10,7 @@ import Withdraw from "./Withdraw";
 import YourPosition from "./YourPosition";
 import YourWallet from "./YourWallet";
 
-export type Method = "create" | "deposit" | "withdraw" | "redeem" | "transfer";
```
mrice32, line 13: nice. This is definitely more of a distraction than anything else right now.

features/manage-position/Create.tsx
nicholaspai: It' just a space. Could be equivalen tot [`CR of newly created tokens: ${value}`]

features/all-positions/AllPositions.tsx
```diff
@@ -61,80 +59,114 @@ const AllPositions = () => {
...
   ) => {
-    if (!latestPrice) return null;
+    if (latestPrice === 0) return null;
     const tokensScaled = Number(tokens) * Number(latestPrice);
     return (Number(collateral) / tokensScaled).toFixed(4);
   };
 
-  return (
-    <Box>
+  if (
```
nicholaspai, line 67: @chrismaree @mrice32 I think this is a pretty clean and organized way to handle the the null-check for rendering logic (#68) and I'm going to apply it across all components in this PR. Here's the idea:

Import and load all containers first, and then define any helper functions.

Only render the main page if all "data" from containers and other related variables are deemed valid (i.e. not `null` or `undefined`). This results in the big `if` statement before the main render `return()`.

If any of the "data" is not ready, then show the fallback JSX component. This way all of the main rendering logic can assume that data is valid.

features/contract-state/GeneralInfo.tsx
```diff
@@ -44,101 +44,127 @@ const GeneralInfo = () => {
...
-  return (
-    <Box>
-      <Typography variant="h5">
-        General Info{" "}
-        {contract?.address && (
-          <Link
-            href={getEtherscanUrl(contract.address)}
-            target="_blank"
-            rel="noopener noreferrer"
+  if (
```
nicholaspai, line 47: @mrice32 @chrismaree @adrianmcli 

I think I also came up with a way to display "stateless" components and maximize "discoverability. 

If you scroll down to the end of this file you'll see a function `renderComponent` that takes in data to display. This is the stateless component and should work even if the data is "missing" because we define default values for it.

Back to this `if` statement logic: We will set a series of data variables to display IFF all of the data is valid and in the expected form. If all of the data is ready then we'll pass on the overidden values to `renderComponent()` otherwise we'll just call `renderComponent()` with no arguments and let the defaults carry over.

PR #66, opened 2020-07-24
Support more wallets with Onboard.js

chrismaree:
This PR adds Onboard.j. These add a _much_ wider range of web3 connectors and gives transaction tracking. The current config works on mainnet & Kovan.

Wallet connector on desktop 
image

Wallet connector on mobile
image

Pending transactions on main net:
image
image

Note that in the current implementation Notify.js will break if running on kovan. It detects the transaction has been sent but does not proceed to notify when the transaction has `succeeded`. Other than the network switching on notify.js this implementation works perfectly between networks.

I have validated the behavior of this PR by making a main net position and doubled checked the correct notifications were generated.

Close: #82


features/manage-position/Withdraw.tsx
```diff
@@ -440,4 +445,4 @@ const Deposit = () => {
   }
 };
 
-export default Deposit;
+export default Withdraw;
```
chrismaree, line 448: unrelated to this PR but this was a typo we've had for a while

containers/Connection.ts
```diff
@@ -1,54 +1,116 @@
...
   const [address, setAddress] = useState<string | null>(null);
+  const [selectedWallet, setSelectedWallet] = useState<string | null>(null);
   const [error, setError] = useState<Error | null>(null);
   const [block$, setBlock$] = useState<Observable<Block> | null>(null);
 
   const attemptConnection = async () => {
-    if (window.ethereum === undefined) {
-      throw Error("MetaMask not found, please visit https://metamask.io/");
-    }
+    const apiKey = process.env.REACT_APP_ONBOARD_API_KEY
```
chrismaree, line 28: note the dapp will work fine without specifying any of these params. They default to common public values (public Infura key & the default onboard.js API key)

containers/Collateral.ts
```diff
@@ -51,7 +51,11 @@ function useCollateral() {
   const setMaxAllowance = async () => {
     if (contract && empAddress) {
       try {
-        await contract.approve(empAddress, ethers.constants.MaxUint256);
+        const tx = await contract.approve(
+          empAddress,
+          ethers.constants.MaxUint256
+        );
+        notify?.hash(tx.hash);
```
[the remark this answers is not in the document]
nicholaspai: I think it sends a [notification](https://docs.blocknative.com/notify)

containers/Connection.ts
nicholaspai: Going off matt's questions, who's keys are these and why would you ever want to set your own keys?

containers/Connection.ts
```diff
@@ -1,54 +1,112 @@
...
+            walletName: "imToken",
+            rpcUrl:
+              !!network && network.chainId === 1
+                ? "https://mainnet-eth.token.im"
+                : "https://eth-testnet.tokenlon.im",
+            preferred: true,
+          },
+          { walletName: "coinbase", preferred: true },
+          {
+            walletName: "portis",
```
nicholaspai, line 72: When I tried to test this out I only got Metamask option

containers/Connection.ts
```diff
@@ -1,54 +1,112 @@
...
+            const ethersProvider = new ethers.providers.Web3Provider(
+              wallet.provider
+            );
+            setProvider(ethersProvider);
+            setNetwork(await ethersProvider.getNetwork());
+            setSigner(ethersProvider.getSigner());
+            await ethersProvider.send("eth_requestAccounts", []);
+            setAddress(await ethersProvider.getSigner().getAddress());
+            setSelectedWallet(wallet.name);
+          } else {
```
nicholaspai, line 54: I think you should set the Provider and all of these values to null together. Is it possible for network, signer, address, etc. to be non null if provider `wallet.provider` doesn't exist?

features/manage-position/Create.tsx
```diff
@@ -143,6 +145,7 @@ const Create = () => {
           const tokensWei = toWei(tokens);
           const tx = await emp.create([collateralWei], [tokensWei]);
           setHash(tx.hash as string);
+          notify?.hash(tx.hash);
```
nicholaspai, line 148: Should we remove the the transaction hash/success/error component from this and other files now?

features/manage-position/Withdraw.tsx
nicholaspai: Lol woops

adrianmcli:
This looks like a great UX improvement. I'm not familiar w/ the Onboard API, but it seems like things are laid out alright.

containers/Connection.ts
chrismaree: This is the sample key from the onboard.js repo. note that not all wallets need this param. Metamask, for example, does not require you to provide this key. 

I'm not actually sure how the dos protection would work in this case. I mean if it's set as a .env variable for the server then the risk is pretty low as no one can actually find it and dos it. In the official docs it's just that some wallet providers require this as a field 🤷‍♂️ https://docs.blocknative.com/onboard#built-in-modules
I can reach out to them and ask about this specific point as I'm not actually sure.

containers/Collateral.ts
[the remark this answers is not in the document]
chrismaree: yup. wherever we generate a hash you pass it to the `notify` object and it will track it until it enters a block and provides notifications to the user.

containers/Connection.ts
chrismaree: done with setting everything else to null.

I'm not sure about how it handles edge cases. need to test that a bit more.

containers/Connection.ts
chrismaree: really? can you post a screenshot? My side looks like this on desktop:
image

containers/Connection.ts
```diff
@@ -1,54 +1,112 @@
...
+            walletName: "walletConnect",
+            rpc: { [network?.chainId || 1]: infuraRpc },
+          },
+          { walletName: "walletLink", rpcUrl: infuraRpc },
+          { walletName: "opera" },
+          { walletName: "operaTouch" },
+          { walletName: "torus" },
+          { walletName: "status" },
+          { walletName: "unilogin" },
+          // { walletName: "authereum" },
```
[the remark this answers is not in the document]
chrismaree: I've seen reports that it can be a bit dodgy so removed it for now.

features/manage-position/Create.tsx
chrismaree: I thought about that. I dont really see any reason to leaving them both for now. Can strip them out if need be but it's useful to get both I think

containers/Collateral.ts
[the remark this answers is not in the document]
nicholaspai: What happens if there are multiple transactions pending at the same time for the user?

containers/Connection.ts
nicholaspai: Is there any way to avoid having these env vars in the dApp? They're not very clean and I'd like to avoid having these .env dependencies even if they're not a problem in run-time.

containers/Connection.ts
```diff
@@ -1,54 +1,116 @@
...
+      },
+      walletSelect: {
+        wallets: [
+          { walletName: "metamask", preferred: true },
+          {
+            walletName: "imToken",
+            rpcUrl:
+              !!network && network.chainId === 1
+                ? "https://mainnet-eth.token.im"
+                : "https://eth-testnet.tokenlon.im",
```
nicholaspai, line 71: Can you explain this? "imToken" uses two servers, one for mainnet and one for everything else?

containers/Connection.ts
```diff
@@ -1,54 +1,116 @@
...
+            preferred: true,
+          },
+          { walletName: "coinbase", preferred: true },
+          {
+            walletName: "portis",
+            apiKey: process.env.REACT_APP_PORTIS_API_KEY,
+          },
+          { walletName: "trust", rpcUrl: infuraRpc },
```
nicholaspai, line 79: So TrustWallet, Dapper, WalletLink, all require an infura node to work? I thought they would have their own provider to iinject?

features/weth/Weth.tsx
```diff
@@ -81,7 +81,7 @@ const Weth = () => {
     wethBalance,
     ethBalance,
   } = WethContract.useContainer();
-  const { signer } = Connection.useContainer();
+  const { signer, notify } = Connection.useContainer();
```
nicholaspai, line 84: This file doesn't look like its merged in latest with master, its prob worth rebasing now and fix some conflicts

nicholaspai:
@chrismaree also your images of wallet provider displays don't work can you reattach?

containers/Connection.ts
chrismaree: well this is why it's optional and there are defaults :) if you don't provide them the app still works fine but ideally, you want to have them.

Onboard.js, if provided with your own key, is pretty useful. it gives metrics of the users that connect their wallets and statistics around them.

chrismaree: I do agree though that this is the main downside with onboard.js. If we feel really strongly about this we can try a different provider, like [web3modal](https://github.com/Web3Modal/web3modal). I do think though that some providers will _always_ need some kind of keys, such as portis/fortmatic/ImToken.

containers/Collateral.ts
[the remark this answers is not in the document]
chrismaree: I _think_ it handles this gracefully. I have not tested this specific case.

containers/Connection.ts
chrismaree: ye, I'm token is _really_ hard to work with. their `https://eth-testnet.tokenlon.im` is actually just their own kovan node. it's pretty inconsistent and getting it to work has been a major pain in other projects I've worked on before; we could _never_ get imtoken to connect to anything other than their own providers.

This is not a big issue for us given we are on kovan :)

containers/Connection.ts
chrismaree: ye, it's a bit of a pain but they all need you to give the provider 🤷

chrismaree:
> @chrismaree also your images of wallet provider displays don't work can you reattach?

@nicholaspai do you mean the wallet screenshots in the PR? they load fine for me but here is another screenshot for completeness 
image

image

features/weth/Weth.tsx
chrismaree: done and updated PR to use master

containers/Connection.ts
nicholaspai: Ok I'm good with this and using our own onboard.js key since we can get metrics!

However, do you know how we set configs in Vercel?

containers/Connection.ts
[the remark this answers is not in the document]
nicholaspai: I'd rather just remove it entirely instead of commenting it out

containers/Connection.ts
nicholaspai: OK well if you see it then that's great. I have a metamask extension on chrome and only see metamask option

features/manage-position/Create.tsx
nicholaspai: Ok we can strip out in another PR this one is chunky already

containers/Connection.ts
```diff
@@ -1,54 +1,116 @@
...
   const attemptConnection = async () => {
-    if (window.ethereum === undefined) {
-      throw Error("MetaMask not found, please visit https://metamask.io/");
-    }
-
-    // // for testing
-    // const provider = new ethers.providers.JsonRpcProvider()
-    // let signer = new ethers.Wallet(process.env.PRIV_KEY)
-    // signer = signer.connect(provider)
+    const apiKey = process.env.REACT_APP_ONBOARD_API_KEY
```
nicholaspai, line 28: For better environment maangement, I'd like to see all `process.env.VAR` refactored into their own `config.js` file. From experience, it's much easier if we have one dedicated file to detecting environment variables and/or setting default values. This way in downstream files such as this, we don't need the `VAR ? VAR : DEFAULT` pattern and can just call `VAR`.

WDYT?

containers/Connection.ts
chrismaree: Great. It's pretty [easy](https://vercel.com/blog/environment-variables-ui#:~:text=New%20Project%20Setting,field%20in%20the%20Project%20Settings.) to do env vars in Vercel. I'll set them up now before merging this in.

features/manage-position/Create.tsx
chrismaree: It covers all functions already though. I'm happy to pull it into two PRs so we can test. 

Another option to "test in production" would be to circulate this testing URL internally for anyone creating a transaction and validate that no errors are thrown.

containers/Config.ts
```diff
@@ -0,0 +1,54 @@
...
+  const infuraId =
+    process.env.REACT_APP_INFURA_ID || "d5e29c9b9a9d4116a7348113f57770a8";
+  const infuraRpc = `https://${network?.name}.infura.io/v3/${infuraId}`;
+  const apiKey = process.env.REACT_APP_ONBOARD_API_KEY
+    ? process.env.REACT_APP_ONBOARD_API_KEY
+    : "12153f55-f29e-4f11-aa07-90f10da5d778";
+  return {
+    apiKey: apiKey ? apiKey : "12153f55-f29e-4f11-aa07-90f10da5d778",
+    portisKey: process.env.REACT_APP_PORTIS_API_KEY,
+    onboardWalletSelect: {
```
nicholaspai, line 13: nit: I'd recommend you put the `apiKey`, `portisKey`, `onboardWalletSelect` and `wallets` into a `portisConfig` key. I think we'll want to use this container for other external SDK's in the future and the name `apiKey` will be reused a lot

containers/Config.ts
```diff
@@ -0,0 +1,54 @@
+import { ethers } from "ethers";
+type Network = ethers.providers.Network;
+export const config = (network: Network | null) => {
```
nicholaspai, line 3: I like passing in `network` here because you could imagine all of the config changing with a diff network

containers/Connection.ts
```diff
@@ -1,54 +1,73 @@
...
 import { useState, useEffect } from "react";
 import { ethers } from "ethers";
+import { API as OnboardApi, Wallet } from "bnc-onboard/dist/src/interfaces";
+import { API as NotifyApi } from "bnc-notify/dist/src/interfaces";
+import Onboard from "bnc-onboard";
+import Notify from "bnc-notify";
 import { Observable } from "rxjs";
 import { debounceTime } from "rxjs/operators";
 
+import { config } from "./Config";
```
nicholaspai, line 11: If you wrap all the onboard related stuff into an `onboard` key you can then do `const { onboard } = config`,

and then easily access `dappId: onboard.apiKey`

containers/Config.ts
nicholaspai: Another nit: Should these all be CAPITALIZEd variable names since they're all hardcoded values?

i.e.

API_KEY, PORTIS_KEY, ONBOARD_WALLET_SELECT...

containers/Config.ts
chrismaree: ye, I was thinking we could re-use this file for addresses even for network switching (I know there is `EMPs.ts` but this could join all configs in one place for ease of access.

features/manage-position/Create.tsx
chrismaree: I have removed notify.js to put into a separate PR. Cleaner this way.

containers/Config.ts
chrismaree: I have refactored this accordingly. PTAL. 

I'm not a fan of the all caps on the hardcoded vars though. Also this is a function that can return different valued based on the context. [Airbnb](https://github.com/airbnb/javascript#naming--uppercase) standard says that uppercase constants should only be used if `the programmer can trust it (and its nested properties) to never change` but this is not really the case as the config changes with network param.

containers/Connection.ts
chrismaree: I don't think the `onboard` object should go in the config. it's more part of `Connection.ts`'s context, where it is right now. 

If we were to move it then it should perhaps go to a `onboard.ts` file that `Connection.ts` could import. This `onboard.js` would also use the `config.ts` file. However, I think this is an over-engineered solution and leaving it where it is now in `Connection.ts` is simple enough while still having sufficient separation of concerns.

nicholaspai:
Actually I agree. Makes sense!

containers/Connection.ts
```diff
@@ -1,54 +1,64 @@
...
-    // let signer = new ethers.Wallet(process.env.PRIV_KEY)
-    // signer = signer.connect(provider)
-
-    // get provider and signer
-    const provider = new ethers.providers.Web3Provider(
-      window.ethereum as ExternalProvider
-    );
-    const signer = provider.getSigner();
-    const network = await provider.getNetwork();
+    const onboard = Onboard({
```
nicholaspai, line 27: Wait, why do you declare `onboard` here but also in the `useState()` on line 18?

containers/Connection.ts
```diff
@@ -77,7 +87,18 @@ function useConnection() {
     }
   }, [provider]);
 
-  return { provider, signer, network, address, connect, error, block$ };
+  return {
+    provider,
+    onboard,
+
```
nicholaspai, line 93: Random space here?

containers/Connection.ts
chrismaree: the reason is that I need to full set up the onboard instance with `await onboardInstance.walletSelect();` before setting into the container hook. I have re-named this variable to make it a bit clear that this is the `onboardInstance`..

Issue #68, opened 2020-07-24
Get rid of null checks in rendering logic


mrice32:
I was thinking about how we might avoid the null checks in normal logic throughout the dapp. In the past I've generally used loading guards to block off the logic until the data is all non-null.

However, @adrianmcli mentioned is that the user shouldn't have to connect their wallet to view read-only info. I like that philosophy a lot, but there are a few issues with it in this dapp right now:
- Many of these pages are't viewable until the EMP-selector has data, so these pages can't be seen without connecting anyway.
- Since we're unable to load much data on the page before connecting, the data that the user can see (at least on many pages) is somewhat useless.

I think we should:
- Put loading guards at the top of most of the pages (at least the ones that require an EMP) to block the render function until _all_ required data is loaded. This would look like:

```js
if (someDataFromContainer === null || otherDataFromContainer === null || ... ) {
  return (
    <LoadingJSX />
  );
}
```

- Figure out how to connect a read-only provider that would allow us to load blockchain data without the user connecting a wallet.

Thoughts @nicholaspai @chrismaree @adrianmcli?

nicholaspai:
I totally agree that the null checks are making the code very unreadable. I think this is a great idea: we will check for `null` or other invalid edge cases up front and then assume from there on that the data is in the valid form. I'll start implementing this on #55 for the Create/Withdraw/Redeem/Deposit features

PR #71, opened 2020-07-26
Refactor invalid container checking for WETH.tsx

nicholaspai:
Follows up on the style introduced in #55 to clean up how rendering logic checks invalid data.

Part of #68 . Following this PR, the last file to refactor will be YieldCalculator.tsx


features/weth/Weth.tsx
```diff
@@ -114,33 +132,199 @@ const Weth = () => {
...
-  };
+      switch (action) {
+        case ACTION_TYPE.WRAP:
+          _submitTxn(ethAmountToWrap, ACTION_TYPE.WRAP);
+          break;
+        case ACTION_TYPE.UNWRAP:
+          _submitTxn(wethAmountToUnwrap, ACTION_TYPE.UNWRAP);
+          break;
+        default:
+          alert("Invalid action!");
```
chrismaree, line 150: Dont feel super strongly but just a question of style here: do we need to have this switch logic like this? the `_submitTxn` function seems a bit over-engineered a bit. Could you not simply deal with the two cases in two separate functions and then call the respective functions from the `wrap` and `unwrap` buttons?

features/weth/Weth.tsx
```diff
@@ -114,33 +132,199 @@ const Weth = () => {
...
+                    onSubmit={(e) => handleClick(e, ACTION_TYPE.UNWRAP)}
+                  >
+                    <TextField
+                      fullWidth
+                      type="number"
+                      label="WETH"
+                      value={wethAmount}
+                      onChange={(e) => setWethAmount(e.target.value)}
+                      variant="outlined"
+                      inputProps={{ min: "0", max: wethBalance as string }}
```
chrismaree, line 254: +1 on setting the max to the ethBalance. that's pretty useful. One question though: do we want to restrict this input field? I agree 100% with disabling the button but from a ux perspective do you want to _block_ the user from interacting with this field? might be cleaner to let them enter any wrap/unwrap button but if it's bigger than their balance then just let it error out with the `error` modifier and `helperText`. Therefore my suggestion is to remove this `inputProps max` field.

features/weth/Weth.tsx
```diff
@@ -114,33 +132,199 @@ const Weth = () => {
...
+                            </Button>
+                          </InputAdornment>
+                        ),
+                      }}
+                    />
+                  </form>
+                </InputElement>
+                <InputElement>
+                  <form
+                    style={{ width: "100%" }}
```
chrismaree, line 242: there is a `fullwidth` prop that you should be able to apply. does that not work for the `form` ?

features/weth/Weth.tsx
nicholaspai: You could but is that better style to have 2 functions instead of 1 with a switch?

features/weth/Weth.tsx
nicholaspai: Yeah its not a property of <form>. I think its a Material UI prop right and this is just a plain form

features/weth/Weth.tsx
nicholaspai: Why would a user ever want to wrap/unwrap more than their balance? it would never succeed right?

features/weth/Weth.tsx
```diff
@@ -84,27 +86,43 @@ const Weth = () => {
...
 
-  const _submitTxn = async (amount: string | null, action: ACTION_TYPE) => {
-    if (weth) {
-      if (amount && utils.parseEther(amount).gt(0)) {
+  if (
+    weth !== null &&
+    wethBalance !== null &&
+    ethBalance !== null &&
+    signer !== null
+  ) {
```
mrice32, line 100: nit: thoughts on just doing an early return if these are null? Then all the happy path logic wouldn't be nested and this might be slightly easier to follow.

Something like:
```
if (weth === null && ... ) {
  // Return loading component
}

// More complicated happy path logic
```

features/weth/Weth.tsx
chrismaree: I'm saying from a UX perspective it's restrictive to block the user and I'd rather let them enter an invalid value then tell them it's above their balance. I'm not suggesting letting them submit the tx; the button should be disabled.

However, the pattern you've got here is the same as the rest of the app so it's ok. It just felt counter-intuitive when I played with it on the test link

features/weth/Weth.tsx
chrismaree: well you end up having two functions anyways with `handelClick` and `_submitTxn` :)

 `handelClick` has a switch case in it and `_submitTxn` uses an if then else. if you had 1 function `wrapEth` and another `unwrapWeth` you would not need any switching or if/else logic within them and just the exact calls to the contracts with tx handling. Line 205 would then simply call `wrapEth`. 

I know it's the same as what your current logic does I just stylistically think you could do the same logic with about 50% as much code. 

Ultimately I don't really mind too much, just prefer to have simpler functions with no branching logic.

nicholaspai: ok I'll just create wrapEth and unwrapWeth I think that makes sense

features/weth/Weth.tsx
nicholaspai: There is an early return in the `else` component of this if statement, its just buried at the bottom of the file. I could technically reverse the boolean check to `if(condition === null || condition === null)...` which would have the early return show up earlier in the file but this is how I set it up with rest of repo

PR #89, opened 2020-07-28
Refine YUSD yield calculator

chrismaree:
This PR refines the yield calculator page by adding a mining yield calculator and adds a "buyer" and "seller" switch on yUSD yield. 

image

image


features/yield/FarmingCalculator.tsx
```diff
@@ -0,0 +1,172 @@
...
+    }
+    if (!pool) {
+      return null;
+    }
+    if (!umaPrice) {
+      return null;
+    }
+    const totalValueAddedToPool =
+      Number(USDCAdded) + Number(yUSDAdded) * Number(usdPrice);
+    if (!totalValueAddedToPool || totalValueAddedToPool == 0) {
```
nicholaspai, line 51: I think we should keep all these null or otherwise invalid data checks as close to the top of the function as possible for readability. If anything, I'd recommend combining them all into one if statement

features/yield/FarmingCalculator.tsx
```diff
@@ -0,0 +1,172 @@
...
+      <Box pt={2}>
+        <form noValidate autoComplete="off">
+          <Grid container spacing={2}>
+            <Grid item md={4} sm={6} xs={12}>
+              <FormInput>
+                <TextField
+                  fullWidth
+                  type="number"
+                  label="yUSD added to pool"
+                  value={yUSDAdded?.toString() || ""}
```
nicholaspai, line 101: You should follow the pattern we now have in every other component where we refactor all the null checking to the front of the function and don't have to do it at all in the JSX components

features/yield/FarmingCalculator.tsx
```diff
@@ -0,0 +1,172 @@
...
+                />
+              </FormInput>
+            </Grid>
+            <Grid item md={4} sm={6} xs={12}>
+              <FormInput>
+                <TextField
+                  fullWidth
+                  type="number"
+                  label="USDC added to pool"
+                  value={USDCAdded?.toString() || ""}
```
nicholaspai, line 117: Same comment as above

features/yield/YieldCalculator.tsx
```diff
@@ -24,10 +33,34 @@ const YieldCalculator = () => {
   const { empState } = EmpState.useContainer();
   const { expirationTimestamp } = empState;
   const { usdPrice } = Balancer.useContainer();
-
   const [tokenPrice, setTokenPrice] = useState<string | null>(
     usdPrice ? usdPrice.toString() : null
   );
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+  const [selectedUserMode, setSelectedUserMode] = useState<string | null>(
```
nicholaspai, line 40: I think you should use a radio button instead of a menu, it reduces # of clicks

pages/index.tsx
```diff
@@ -96,7 +96,6 @@ export default function Index() {
               </Grid>
             </Box>
             <Menu
-              id="lock-menu"
```
nicholaspai, line 99: What does this do?

features/yield/FarmingCalculator.tsx
```diff
@@ -0,0 +1,172 @@
...
+  margin-bottom: 20px;
+`;
+
+const WEEKS_PER_YEAR = 52;
+
+const YieldCalculator = () => {
+  const { empState } = EmpState.useContainer();
+  const { usdPrice, pool } = Balancer.useContainer();
+  const [umaPrice, setUmaPrice] = useState<number | null>(4.0);
+  const [yUSDAdded, setyUSDAdded] = useState<string | null>("1000");
```
nicholaspai, line 21: Also, FYI we can easily query the user's shares and % of total shares in pool if they connect their web3 provider. Maybe we should make farming calculator only available if they connect their web3 provider and then we can just detect their shares amount.

Try this query:
```
{
pools(where: {tokensList: ["0x81ab848898b5ffd3354dbbefb333d5d183eedcb5", "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"]}) {
    shares(first:1000) {
      userAddress {
        id
      }
      balance
    }
  }
}
```

@ https://thegraph.com/explorer/subgraph/balancer-labs/balancer

chrismaree: Ye, This is the query that I used in the LP mining script. might be nice to update the table with their actual amount provided in the same way we update the token price in the yUSD payout calculator. however I think this should perhaps be a separate PR as this is already 300+ lines changed.

nicholaspai: Yeah I think we should think of this PR as setting up the TSX components/container

features/yield/FarmingCalculator.tsx
chrismaree: This i'll leave for a separate PR. good to do all these reactors in one go

features/yield/YieldCalculator.tsx
chrismaree: sure & done :)

pages/index.tsx
chrismaree: typo. have removed

features/yield/FarmingCalculator.tsx
chrismaree: cool. will leave for a leter PR if we deem it wanted

chrismaree:
@nicholaspai please let me know if there is anything else you want me to change, I've replied to all your comments.

PR #95, opened 2020-07-28
Show total number of sponsors in general information

chrismaree:
This small PR adds the total number of sponsors within a given EMP to the general Info tab.
image


features/contract-state/DisputeParams.tsx
```diff
@@ -23,7 +30,11 @@ const DisputeParams = () => {
     disputeBondPct,
     disputerDisputeRewardPct,
     sponsorDisputeRewardPct,
+    withdrawalLiveness,
   } = empState;
+  const withdrawalLivenessInMinutes = (Number(withdrawalLiveness) / 60).toFixed(
```
nicholaspai, line 35: This withdrawal liveness is already shown in general info?

features/contract-state/GeneralInfo.tsx
```diff
@@ -40,11 +41,16 @@ const GeneralInfo = () => {
...
     collateralRequirement: collReq,
     minSponsorTokens,
-    withdrawalLiveness,
   } = empState;
   const { symbol: tokenSymbol } = Token.useContainer();
+
+  const [sponsorCount, setSponsorCount] = useState<string | null>(null);
   const defaultMissingDataDisplay = "N/A";
 
+  useEffect(() => {
```
nicholaspai, line 50: `activeSponsors` is already updated within an `useEffect` hook in the EMPSponsors container, so why not just add `activeSponsors !== null` to the big `if` statement, and then just create a variable `const sponsorCount = Object.keys(activeSponsors).length`? I don't think you need to create the state variable as you did

features/contract-state/GeneralInfo.tsx
```diff
@@ -60,11 +66,9 @@ const GeneralInfo = () => {
     ).toLocaleString("en-GB", { timeZone: "UTC" });
     const prettyLatestPrice = Number(latestPrice).toFixed(4);
     const pricedGcr = (gcr / latestPrice).toFixed(4);
-    const withdrawalLivenessInMinutes = (
-      Number(withdrawalLiveness) / 60
-    ).toFixed(2);
+
     const priceIdUtf8 = parseBytes32String(priceId);
-    const collReqPct = (parseFloat(fromWei(collReq)) * 100).toString();
+    const collReqPct = parseFloat(fromWei(collReq)).toString();
```
nicholaspai, line 71: +1 as its more consistent with how we display CR in other files

features/contract-state/DisputeParams.tsx
chrismaree: nope. I removed it from general and put it in the dispute section. see the screenshot for how it looks now.

features/contract-state/GeneralInfo.tsx
chrismaree: yup, that's a cleaner way of doing it. I didn't know I could access the `activeSponsors` like that from the EMPSponsors container. I have refactored, PTAL.

chrismaree:
I have implemented your nits, please take a look @nicholaspai

features/contract-state/GeneralInfo.tsx
```diff
@@ -1,37 +1,38 @@
 import styled from "styled-components";
 import { utils } from "ethers";
 import { Typography, Box, Tooltip } from "@material-ui/core";
+import { useState, useEffect } from "react";
```
nicholaspai, line 4: You can remove `useState` and `useEffect` now right?

chrismaree: ye, you right! have removed this & pushed up for the merge 🚀

PR #96, opened 2020-07-28
Refactor `getOffchainPrice` to handle additional pricefeed sources

### LINKED ISSUE #80: Add Binance (ETHUSDT) and Kraken (ETHUSD) to pricefeed index
Official pricefeeds introduced in this [UMIP](https://github.com/UMAprotocol/UMIPs/pull/23)

nicholaspai:
Since the list of sources is now > 1, its too bulky to list all sources in multiple components. I removed the list of source exchanges from AllPositions and YourPosition.

I added a `Label` to GeneralInfo to display a list of exchanges in order to save more horizontal space.

Resolves #80

### DIFF
--- a/containers/PriceFeed.ts
+++ b/containers/PriceFeed.ts
@@ -12,7 +12,7 @@ function usePriceFeed() {
   const { symbol: tokenSymbol } = Token.useContainer();
 
   const [latestPrice, setLatestPrice] = useState<number | null>(null);
-  const [source, setSource] = useState<string | undefined>(undefined);
+  const [source, setSource] = useState<string[] | undefined>(undefined);
 
   const queryPrice = async () => {
     setLatestPrice(null);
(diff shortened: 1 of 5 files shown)


nicholaspai:
> This looks awesome! is there any latency implications if one of the feeds are slower than others? if one feed goes down does everything continue to work as expected?

As long as one price returns it will work, otherwise returns null like it does now if no coinbase price. All promises are resolved in parallel: 

```
const prices: (number | null)[] = await Promise.all(
      identifierParams.source.map(async (url: string) => {
        try {
          const response = await fetch(url);
          const json = await response.json();

          switch (true) {
            case url.includes("coinbase"):
              return _getCoinbasePriceFromJSON(json);
            case url.includes("binance"):
              return _getBinancePriceFromJSON(json);
            case url.includes("kraken"):
              return _getKrakenPriceFromJSON(json);
            default:
              return null;
          }
        } catch (err) {
          console.error(
            `Failed to get price for for token with symbol ${symbol}, url=${url}`,
            err
          );
          return null;
        }
      })
```

PR #100, opened 2020-07-29
implemented eip747

chrismaree:
This PR adds a simple button to add the synthetic token to your metamask wallet. Additionally, this PR hides etherscan & market links until the user has selected a contract from the dropdown list.
image

image

Open question: do we want this functionality? main issue is the token is limited to 6 chars in eip747. This is pretty painful as we loose token information (see screenshot).

### DIFF
--- a/features/contract-state/Totals.tsx
+++ b/features/contract-state/Totals.tsx
@@ -1,11 +1,12 @@
 import styled from "styled-components";
-import { Box, Grid, Typography } from "@material-ui/core";
+import { Box, Grid, Typography, Button } from "@material-ui/core";
 
 import TotalsContainer from "../../containers/Totals";
 import Collateral from "../../containers/Collateral";
 import Token from "../../containers/Token";
 import Etherscan from "../../containers/Etherscan";
 
+import Connection from "../../containers/Connection";
 import { getExchangeInfo } from "../../utils/getExchangeLinks";
 
 const DataBox = styled(Box)`
@@ -42,6 +43,7 @@ const White = styled.span`
 `;
 
 const Totals = () => {
+  const { provider } = Connection.useContainer();
   const { totalCollateral, totalTokens } = TotalsContainer.useContainer();
   const {
     symbol: collSymbol,
@@ -87,10 +89,29 @@ const Totals = () => {
     prettyTotalTokens: string = defaultMissingDataDisplay,
     prettyCollSymbol: string = "",
     prettyTokenSymbol: string = "",
-    getExchangeLinkCollateral: string = "https://app.uniswap.org/#/swap",
-    getExchangeLinkToken: string = "https://app.uniswap.org/#/swap",
+    getExchangeLinkCollateral: string = "",
+    getExchangeLinkToken: string = "",
     exchangeName: string = "Uniswap"
   ) {
+    const addTokenToMetamask = async () => {
+      // Add token to users metamask wallet.
+      if (provider == null) return;
+
+      // @ts-ignore
+      provider.send("wallet_watchAsset", {
+        // @ts-ignore
+        type: "ERC20",
+        options: {
+          address: tokenAddress,
+          symbol: tokenSymbol?.substring(0, 6),
+          name: "test",
+          decimals: 18,
+          image:
+            "https://etherscan.io/token/images/yusdsynthetictokenexpiring1september2020_32.png",
+        },
+        id: Math.round(Math.random() * 100000),
+      });
+    };
     return (
       <>
         <Grid item md={6} xs={12}>
@@ -103,20 +124,24 @@ const Totals = () => {
               of <White>collateral</White> supplied
             </Label>
             <LinksContainer>
-              <SmallLink
-                href={getEtherscanUrl(collAddress)}
-                target="_blank"
-                rel="noopener noreferrer"
-              >
-                Etherscan
-              </SmallLink>
-              <SmallLink
-                href={getExchangeLinkCollateral}
-                target="_blank"
-                rel="noopener noreferrer"
-              >
-                {exchangeName}
-              </SmallLink>
+              {collAddress && (
+                <SmallLink
+                  href={getEtherscanUrl(collAddress)}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  Etherscan
+                </SmallLink>
+              )}
+              {getExchangeLinkCollateral != "" && (
+                <SmallLink
+                  href={getExchangeLinkCollateral}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  {exchangeName}
+                </SmallLink>
+              )}
             </LinksContainer>
           </DataBox>
         </Grid>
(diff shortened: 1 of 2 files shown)


features/contract-state/Totals.tsx
```diff
@@ -103,20 +124,24 @@ const Totals = () => {
...
-                Etherscan
-              </SmallLink>
-              <SmallLink
-                href={getExchangeLinkCollateral}
-                target="_blank"
-                rel="noopener noreferrer"
-              >
-                {exchangeName}
-              </SmallLink>
+              {collAddress && (
```
nicholaspai, line 127: You should pass in `collAddress`  into `renderComponent` so that this file is more functional. And you should check `collAddress !== null` like you do with `getExchangeLinkColalteral` because it is a string type.

features/contract-state/Totals.tsx
```diff
@@ -103,20 +124,24 @@ const Totals = () => {
...
+                <SmallLink
+                  href={getEtherscanUrl(collAddress)}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  Etherscan
+                </SmallLink>
+              )}
+              {getExchangeLinkCollateral != "" && (
```
nicholaspai, line 136: nit: need a third `!==`

features/contract-state/Totals.tsx
```diff
@@ -131,20 +156,31 @@ const Totals = () => {
...
-                Etherscan
-              </SmallLink>
-              <SmallLink
-                href={getExchangeLinkToken}
-                target="_blank"
-                rel="noopener noreferrer"
-              >
-                {exchangeName}
-              </SmallLink>
+              {tokenAddress && (
```
nicholaspai, line 159: Same comments as `collAddress`

features/contract-state/Totals.tsx
```diff
@@ -131,20 +156,31 @@ const Totals = () => {
...
+                <SmallLink
+                  href={getEtherscanUrl(tokenAddress)}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  Etherscan
+                </SmallLink>
+              )}
+
+              {getExchangeLinkToken != "" && (
```
nicholaspai, line 169: Same comment, need a !==

features/contract-state/Totals.tsx
```diff
@@ -131,20 +156,31 @@ const Totals = () => {
...
+                <SmallLink
+                  href={getExchangeLinkToken}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  {exchangeName}
+                </SmallLink>
+              )}
+
+              {tokenAddress && (
```
nicholaspai, line 179: Also should check that `tokenAddress !== ""` to be typesafe

features/contract-state/Totals.tsx
```diff
@@ -87,10 +89,29 @@ const Totals = () => {
     prettyTotalTokens: string = defaultMissingDataDisplay,
     prettyCollSymbol: string = "",
     prettyTokenSymbol: string = "",
-    getExchangeLinkCollateral: string = "https://app.uniswap.org/#/swap",
-    getExchangeLinkToken: string = "https://app.uniswap.org/#/swap",
+    getExchangeLinkCollateral: string = "",
+    getExchangeLinkToken: string = "",
     exchangeName: string = "Uniswap"
   ) {
+    const addTokenToMetamask = async () => {
```
nicholaspai, line 96: You should add `provider !== null` to the early `if` statement checking for null values, and then declare `addTokenToMetamask` within that if statement.

Then you should pass `addTokenToMetamask` as a function into `renderComponent`. This would be consistent with the repo style

features/contract-state/Totals.tsx
```diff
@@ -87,10 +89,29 @@ const Totals = () => {
...
+      // Add token to users metamask wallet.
+      if (provider == null) return;
+
+      // @ts-ignore
```
nicholaspai, line 100: Should this be an async call? I mention because you declare `addTokenToMetamask` as an async func

features/contract-state/Totals.tsx
```diff
@@ -87,10 +89,29 @@ const Totals = () => {
...
+      provider.send("wallet_watchAsset", {
+        // @ts-ignore
+        type: "ERC20",
+        options: {
+          address: tokenAddress,
+          symbol: tokenSymbol?.substring(0, 6),
```
nicholaspai, line 106: If you put this within the `if` statement above you will not need to have the `?` after `tokenSymbol` because you will already have checked for that

features/contract-state/Totals.tsx
chrismaree: I can do that. if we make that change then we should do the same for `collAddress` as well as `tokenAddress` to be consistent. I did not do that with either to be consistent with the previous implementation but am happy to add it.

features/contract-state/Totals.tsx
chrismaree: given this is now passed in from the `renderComponent` with a default string I agree should be checked against an empty string. Have updated it.

features/contract-state/Totals.tsx
chrismaree: I _can_ do that but this feels like it deviates from the pattern tbh. The `renderComponent` logic is not used to pas around functions like this in other places.

chrismaree: I implemented the changes you suggested. take a look and let me know if this is what you meant

features/contract-state/Totals.tsx
chrismaree: it turns out it does not need to be async. I have removed it accordingly

features/contract-state/Totals.tsx
chrismaree: great, good catch

chrismaree:
@nicholaspai  I've implemented your nits. One thing I'm not really sure about is if we actually want to use the `renderComponent` to feed in functions like this. We don't do this anywhere else in the repo. Can you clarify exactly where you want to use this pattern and were we don't? I thought we only wanted to do that for displayed properties.

features/contract-state/Totals.tsx
nicholaspai: Yeah it's easy enough to pass in and make the change to be consistent

features/contract-state/Totals.tsx
```diff
@@ -69,14 +72,37 @@ const Totals = () => {
     const getExchangeLinkToken = exchangeInfo.getExchangeUrl(tokenAddress);
     const exchangeName = exchangeInfo.name;
 
+    const addTokenToMetamask = () => {
+      // Add token to users metamask wallet.
+      if (provider == null) return;
```
nicholaspai, line 77: So you can remove this `provider === null` check now that you already check it right?

features/contract-state/Totals.tsx
```diff
@@ -87,9 +113,12 @@ const Totals = () => {
...
     prettyTokenSymbol: string = "",
-    getExchangeLinkCollateral: string = "https://app.uniswap.org/#/swap",
-    getExchangeLinkToken: string = "https://app.uniswap.org/#/swap",
-    exchangeName: string = "Uniswap"
+    collAddress: string = "",
+    tokenAddress: string = "",
+    getExchangeLinkCollateral: string = "",
+    getExchangeLinkToken: string = "",
+    exchangeName: string = "Uniswap",
+    addTokenToMetamask: any = null
```
nicholaspai, line 121: I wonder if instead of `any` you could just make it of type `() => void;`

nicholaspai:
@chrismaree metacomment- who created the yUSD token symbol?

chrismaree:
> @chrismaree metacomment- who created the yUSD token symbol?

ye, I don't know and it looks pretty terrible. We can get a higher quality one for sure. Perhaps should actually upgrade this quality before merging this PR.

adrianmcli:
There were a few good icons proposed in the discord

chrismaree:
> There were a few good icons proposed in the discord

What is the canonical icon? which did we pick?

chrismaree:
> > There were a few good icons proposed in the discord
> 
> What is the canonical icon? which did we pick?

Have chosen an implemented a new icon. Looks like this now:
image

features/contract-state/Totals.tsx
nicholaspai: @chrismaree nit here

features/contract-state/Totals.tsx
chrismaree: this threw an error when I tried. only config I could get ts to pass with is the code here.

PR #101, opened 2020-07-29
SettleExpired Feature

### LINKED ISSUE #102: Query resolved price via UMA subgraph, add to SettleExpired feature

nicholaspai:
If a contract is expired, the user will only see the "Settle" method option. Otherwise, they will see "Create/Deposit/Redeem/Withdraw" but no "Settle"

It allows user to call `expire()` if a contract has expired but has not requested a price yet, and allows either a sponsor or a tokenholder to call `settleExpired()`. The dApp calls `Voting.hasPrice` in order to determine whether a price request has been resolved. Once a price has been resolved, a button to settle expired will appear.

It also provides a Max-Approve button for token holders.

Resolves #102

### DIFF
--- a/containers/DvmState.ts
+++ b/containers/DvmState.ts
@@ -2,11 +2,14 @@ import { createContainer } from "unstated-next";
 import { useState, useEffect } from "react";
 import { utils } from "ethers";
 
+import { useQuery } from "@apollo/client";
+import { PRICE_REQUESTS } from "../apollo/uma/queries";
+
 import Connection from "./Connection";
 import DvmContract from "./DvmContract";
 import EmpState from "./EmpState";
 
-const { formatUnits: fromWei } = utils;
+const { formatUnits: fromWei, parseBytes32String: hexToUtf8 } = utils;
 
 interface ContractState {
   hasEmpPrice: boolean | null;
@@ -26,6 +29,13 @@ enum PRICE_REQUEST_STATUSES {
   FUTURE, // Is scheduled to be voted on in a future round.
 }
 
+interface PriceRequestQuery {
+  identifier: {
+    id: string;
+  };
+  time: string;
+}
+
 const useContractState = () => {
   const { block$ } = Connection.useContainer();
   const { contract: dvm } = DvmContract.useContainer();
@@ -33,6 +43,15 @@ const useContractState = () => {
   const { priceIdentifier, expirationTimestamp } = empState;
 
   const [state, setState] = useState<ContractState>(initState);
+
+  // Because apollo caches results of queries, we will poll/refresh this query periodically.
+  // We set the poll interval to a very slow 5 seconds for now since the position states
+  // are not expected to change much.
+  // Source: https://www.apollographql.com/docs/react/data/queries/#polling
+  const { loading, error, data } = useQuery(PRICE_REQUESTS, {
+    context: { clientName: "UMA" },
+    pollInterval: 5000,
+  });
 
   // get state from EMP
   const queryState = async () => {
@@ -53,21 +72,40 @@ const useContractState = () => {
       const hasPrice = res[0][0].status === PRICE_REQUEST_STATUSES.RESOLVED;
 
       let resolvedPrice = null;
-      if (hasPrice) {
-        try {
-          const postResolutionRes = await Promise.all([
-            // `getPrice()` will fail on mainnet because the deployed `Voting.getPrice()` is required to be called from a registered contract, such as this EMP!
-            // TODO: Figure out a way to spoof the `from` parameter in this contract call and set to `emp.address`. One possibility is via the UMA subgraph
-            dvm.getPrice(
-              priceIdentifier.toString(),
-              expirationTimestamp.toNumber()
-            ),
-          ]);
 
-          resolvedPrice = parseFloat(fromWei(postResolutionRes.toString()));
-        } catch (err) {
-          console.error(`getPrice failed:`, err);
+      // Fetch price from graph if available.
+      if (error) {
+        console.error(`Apollo client failed to fetch graph data:`, error);
+      } else if (hasPrice && !loading && data) {
+        // Implementation 1: Use subgraph data. This works for mainnet but not Kovan since the UMA subgraph does not currently index Kovan data.
+        const matchingPriceRequest = data.priceRequests.find(
+          (request: PriceRequestQuery) => {
+            return (
+              request.identifier.id === "ETH/BTC" && //hexToUtf8(priceIdentifier)
+              request.time === "1592952397" //expirationTimestamp.toString()
+            );
+          }
+        );
+        if (matchingPriceRequest && matchingPriceRequest.isResolved) {
+          resolvedPrice = parseFloat(
+            fromWei(matchingPriceRequest.price.toString())
+          );
         }
+
+        // // Implementation 2: Call contract. This works for Kovan/MockOracle and is therefore useful for testing, but not mainnet since `getPrice` must be called with the overrideL `{from: emp.address}`.
+        // try {
+        //   const postResolutionRes = await Promise.all([
+        //     // `getPrice()` will fail on mainnet because the deployed `Voting.getPrice()` is required to be called from a registered contract, such as this EMP!
+        //     // TODO: Figure out a way to spoof the `from` parameter in this contract call and set to `emp.address`.
+        //     dvm.getPrice(
+        //       priceIdentifier.toString(),
+        //       expirationTimestamp.toNumber()
+        //     ),
+        //   ]);
+        //   resolvedPrice = parseFloat(fromWei(postResolutionRes.toString()));
+        // } catch (err) {
+        //   console.error(`getPrice failed:`, err);
+        // }
       }
 
       const newState: ContractState = {
--- a/containers/DvmState.ts
+++ b/containers/DvmState.ts
@@ -1,0 +1,102 @@
+import { createContainer } from "unstated-next";
+import { useState, useEffect } from "react";
+import { utils } from "ethers";
+
+import Connection from "./Connection";
+import DvmContract from "./DvmContract";
+import EmpState from "./EmpState";
+
+const { formatUnits: fromWei } = utils;
+
+interface ContractState {
+  hasEmpPrice: boolean | null;
+  resolvedPrice: number | null;
+}
+
+const initState = {
+  hasEmpPrice: null,
+  resolvedPrice: null,
+};
+
+// Taken from Voting.sol implementation.
+enum PRICE_REQUEST_STATUSES {
+  NOT_REQUESTED, // Was never requested.
+  ACTIVE, // Is being voted on in the current round.
+  RESOLVED, // Was resolved in a previous round.
+  FUTURE, // Is scheduled to be voted on in a future round.
+}
+
+const useContractState = () => {
+  const { block$ } = Connection.useContainer();
+  const { contract: dvm } = DvmContract.useContainer();
+  const { empState } = EmpState.useContainer();
+  const { priceIdentifier, expirationTimestamp } = empState;
+
+  const [state, setState] = useState<ContractState>(initState);
+
+  // get state from EMP
+  const queryState = async () => {
+    if (
+      dvm !== null &&
+      priceIdentifier !== null &&
+      expirationTimestamp !== null
+    ) {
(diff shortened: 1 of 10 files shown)


nicholaspai:
This feature is hard to test and I've had to tweak the `DeployEMP.js` script to (1) deploy a contract that is expiring soon, and (2) add a `getPriceRequestStatuses` method to the `MockOracle.sol` contract (and redeploy it) so that I can query request statuses {REQUESTED, ACTIVE, RESOLVED} just like we can with `Voting.sol`. Therefore, this is a test case I ran through

TEST CASE: SETUP
- Create an EMP that expires soon after I deploy it that uses the yCOMP price identifier
- Quickly, create an initial sponsor position for 150 yCOMP collateralized by 50,000 DAI (0x9a...)
- Send 100 yCOMP to another sponsor (0xff7...)

TEST CASE: **Contract has expired, but `expire()` has not been called yet. This is possible from the dApp.**
Screen Shot 2020-07-30 at 00 18 51

TEST CASE: **Price request has been sent, but has not resolved yet.**
Screen Shot 2020-07-30 at 00 21 06

TEST CASE: **I submitted a price of 130 for the COMPUSD identifier @ the contract expiration timestamp.**

TEST CASE: **Holder of 100 yCOMP (who has no position) can redeem 100 yCOMP for (100 * 130 = 13,000) DAI.**
Screen Shot 2020-07-30 at 00 26 01
[Transaction Receipt](https://kovan.etherscan.io/tx/0xb73e606f5091da2fdcd52c9eaf0c5359c8ac41f5a5bc0dbfa5fc541731379258)

TEST CASE: **Sponsor holding 50 yCOMP can redeem 50 yCOMP for (50 * 130 = 6,500) DAI + excess collateral worth 50,000 - (150 * 130) = 30,500 DAI for a total of 37,000 DAI.**
Screen Shot 2020-07-30 at 00 41 41
[Transaction Receipt](https://kovan.etherscan.io/tx/0xa14f61f4875647345ce844a7f82e5ba357562a7c21a222dacb0e2c28c3571edd)

Final state: **No collateral or positions remain in the contract**
Screen Shot 2020-07-30 at 00 42 39

containers/DvmState.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      const hasPrice = res[0][0].status === PRICE_REQUEST_STATUSES.RESOLVED;
+
+      let resolvedPrice = null;
+
+      // Fetch price from graph if available.
+      if (error) {
+        console.error(`Apollo client failed to fetch graph data:`, error);
+      } else if (hasPrice && !loading && data) {
+        // Implementation 1: Use subgraph data. This works for mainnet but not Kovan since the UMA subgraph does not currently index Kovan data.
```
nicholaspai, line 80: - I used the Kovan contracts `.getPrice()` calls to test on Kovan, but also verified that this query parsing works on mainnet using hardcoded, already resolved price request data.

containers/DvmState.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  // get state from EMP
+  const queryState = async () => {
+    if (
+      dvm !== null &&
+      priceIdentifier !== null &&
+      expirationTimestamp !== null
+    ) {
+      const res = await Promise.all([
+        dvm.getPriceRequestStatuses([
```
chrismaree, line 64: So to be clear: you dont need to query the `getPendingRequests` to see if the request is enquired here as you can simple check a combination of the expiration timestamp & if the price has been resolved within the components below?

containers/EmpState.ts
```diff
@@ -92,6 +100,9 @@ const useContractState = () => {
         liquidationLiveness: res[13] as BigNumber,
         withdrawalLiveness: res[14] as BigNumber,
         currentTime: res[15] as BigNumber,
+        isExpired: Number(res[15]) >= Number(res[0]),
```
chrismaree, line 103: +1

features/manage-position/MethodSelector.tsx
```diff
@@ -40,44 +42,70 @@ interface IProps {
...
+          >
+            {_isExpired && (
+              <MenuItem value={"settle"}>
+                <ListItemText
+                  primary="Settle"
+                  secondary="Settle expired tokens."
+                />
+              </MenuItem>
+            )}
+            {!_isExpired && (
```
chrismaree, line 72: could you not wrap all the items below this one that you don't want to render when `!_isExpired` in a `<div> ... </div>` and then only query this `!_isExpired` once?

features/manage-position/SettleExpired.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+    const needsToRequestSettlementPrice = contractState === CONTRACT_STATE.OPEN;
+
+    // Error conditions for calling settle expired:
+    const needAllowance =
+      tokenAllowance !== "Infinity" && tokenAllowance < tokenBalance;
+    const canSettleTokens = tokenBalance > 0 || posTokens > 0;
+
+    // Calculate collateral to receive if price was resolved.
+    let positionTRV: number = -1;
```
chrismaree, line 88: this is a new pattern. why are you setting the defaults to -1?

features/manage-position/SettleExpired.tsx
```diff
@@ -0,0 +1,297 @@
...
+          await tx.wait();
+          setSuccess(true);
+        } catch (error) {
+          console.error(error);
+          setError(error);
+        }
+      } else {
+        setError(
+          new Error(
+            "You have no yUSD in your wallet or your position to settle"
```
chrismaree, line 117: nit: this wont only be yUSD :)

containers/DvmState.ts
```diff
@@ -0,0 +1,140 @@
...
+        resolvedPrice: resolvedPrice,
+      };
+
+      setState(newState);
+    }
+  };
+
+  // get state on setting of contract
+  useEffect(() => {
+    setState(initState);
```
chrismaree, line 122: Do you want to re-set the state like this within the `useEffect`? I _think_ this will result in the state having a short discontinuity to `null` between polls. The result of this is that while the state is updated in the `queryState` function anything that is dependent on this state will flicker to the `null` state and then back to the populated state when the data has loaded.

I think it's fine to just remove this line as the state is populated to the initState when it's created and updated to newState on each loop.

I came across this behavior while adding the final fee to the Dispute params in the front end. I forked off this branch and added the `finalFee` to the dvm's `ContractState`. This is in [this](https://github.com/UMAprotocol/emp-tools/pull/103) PR. Once your PR is merged into master I'll point that PR to master as well.

containers/DvmState.ts
```diff
@@ -0,0 +1,140 @@
...
+            );
+          }
+        );
+        if (matchingPriceRequest && matchingPriceRequest.isResolved) {
+          resolvedPrice = parseFloat(
+            fromWei(matchingPriceRequest.price.toString())
+          );
+        }
+
+        // // Implementation 2: Call contract. This works for Kovan/MockOracle and is therefore useful for testing, but not mainnet since `getPrice` must be called with the overrideL `{from: emp.address}`.
```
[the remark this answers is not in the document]
nicholaspai: I figured out how to do it! WHen I construct the `dvm` contract, instead of passing in the `signer` I can just pass in the `provider`. Without a default signer, you can call a contract and override the `from`!

containers/DvmState.ts
nicholaspai: Yeah I only need to check `hasPrice`, and if there is a price then I can get the actual price. The front end only needs to know if a price exists, if it doesn't then `settleExpired()` will fail.

features/manage-position/MethodSelector.tsx
nicholaspai: You're right!

features/manage-position/SettleExpired.tsx
nicholaspai: you'er right, I ended up replacing these as `let positionTRV: number;`

nicholaspai: Which, I'm surprised you can do with TS

features/manage-position/SettleExpired.tsx
nicholaspai: good catch thanks!

containers/DvmState.ts
nicholaspai: Ah you're right! I was wondering why SettleExpired was flickering haha

nicholaspai:
@chrismaree I responded to your comments. I ended up replacing the `getPriceRequestStatuses` call with `hasPrice` since I figured out how to make calls from the `emp.address`. I also removed the graphql query

features/manage-position/SettleExpired.tsx
nicholaspai: Nevermind, not possible. Set them all to type `number | null` and have to check for null lower down where I render them

containers/DvmState.ts
```diff
@@ -0,0 +1,99 @@
...
+    ) {
+      const res = await Promise.all([
+        dvm.hasPrice(
+          priceIdentifier.toString(),
+          expirationTimestamp.toNumber(),
+          { from: empAddress }
+        ),
+      ]);
+
+      const hasPrice = res[0] as boolean;
```
chrismaree, line 54: could you add a comment here to explain exactly the type returned from the `hasPrice` call?

containers/DvmState.ts
```diff
@@ -0,0 +1,99 @@
...
+      };
+
+      setState(newState);
+    }
+  };
+
+  // get state on setting of contract
+  useEffect(() => {
+    queryState();
+  }, [dvm, priceIdentifier, expirationTimestamp]);
```
chrismaree, line 84: this is great now. very clean 🧽

features/manage-position/ManagePosition.tsx
```diff
@@ -1,24 +1,51 @@
...
+  const [method, setMethod] = useState<Method>(DEFAULT_METHOD);
   const handleChange = (e: React.ChangeEvent<{ value: unknown }>) =>
     setMethod(e.target.value as Method);
 
-  if (signer === null) {
+  if (signer !== null && isExpired !== null) {
+    // Whenever the expiry state changes, check if we should change the default method.
+    useEffect(() => {
+      setMethod(DEFAULT_METHOD);
+      if (isExpired) {
```
chrismaree, line 30: +1

features/manage-position/MethodSelector.tsx
```diff
@@ -40,44 +42,64 @@ interface IProps {
...
+      <Box py={2}>
+        <FormWrapper>
+          <InputLabel id="demo-simple-select-label">Actions</InputLabel>
+          <Select
+            labelId="demo-simple-select-label"
+            value={method}
+            onChange={handleChange}
+            input={<BootstrapInput />}
+          >
+            {_isExpired && [
```
chrismaree, line 64: small styling nit: should this be named wit the `_`? feels like this is the only time we've written code like this in the repo.

containers/DvmState.ts
nicholaspai: It's the same as in `EMPState`, I think its a boolean but with a smart contract call you're never sure I think

```
      const newState: ContractState = {
        expirationTimestamp: res[0] as BigNumber,
        collateralCurrency: res[1] as string, // address
        priceIdentifier: res[2] as Bytes,
        tokenCurrency: res[3] as string, // address
        collateralRequirement: res[4] as BigNumber,
        disputeBondPct: res[5] as BigNumber,
        disputerDisputeRewardPct: res[6] as BigNumber,
        sponsorDisputeRewardPct: res[7] as BigNumber,
        minSponsorTokens: res[8] as BigNumber,
        timerAddress: res[9] as string, // address
        cumulativeFeeMultiplier: res[10] as BigNumber,
        rawTotalPositionCollateral: res[11] as BigNumber,
        totalTokensOutstanding: res[12] as BigNumber,
        liquidationLiveness: res[13] as BigNumber,
        withdrawalLiveness: res[14] as BigNumber,
        currentTime: res[15] as BigNumber,
        isExpired: Number(res[15]) >= Number(res[0]),
        contractState: Number(res[16]),
        finderAddress: res[17] as string, // address
      };

```

features/manage-position/MethodSelector.tsx
nicholaspai: Changed to `contractHasExpired`

PR #103, opened 2020-07-30
Add final fee to DVM state & general info

chrismaree:
This PR adds the final fee (liquidation bond) to the front end & state. This will be useful when finishing the "click to liquidate" PR which asl requires this value.
image

This Pr also makes a small refactor to the general info to show the `Is expired` with the same formatting and style as other components on the General info page:
image

close: #106
close: #107

### DIFF
--- a/containers/DvmContracts.ts
+++ b/containers/DvmContracts.ts
@@ -26,13 +26,10 @@ function useContracts() {
         uma.finder.abi,
         provider
       );
-      const res = await Promise.all([
+      const [votingAddress, storeAddress] = await Promise.all([
         finder.getImplementationAddress(utf8ToHex("Oracle")),
         finder.getImplementationAddress(utf8ToHex("Store")),
       ]);
-
-      const votingAddress = res[0];
-      const storeAddress = res[1];
       // Do not inject a provider into this contract so that we can make calls from the EMP's address.
       // We will only have read-only access to the Contract, but overriding the `from` address is neccessary for `getPrice` or `hasPrice`.
       // Moreover, we won't be submitting any txns to the DVM.
--- a/containers/DvmContracts.ts
+++ b/containers/DvmContracts.ts
@@ -8,7 +8,7 @@ import Connection from "./Connection";
 
 const { formatBytes32String: utf8ToHex } = ethers.utils;
 
-function useContract() {
+function useContracts() {
   const { provider } = Connection.useContainer();
   const { empState } = EmpState.useContainer();
   const { finderAddress } = empState;
@@ -57,6 +57,6 @@ function useContract() {
   return { votingContract, storeContract };
 }
 
-const Contract = createContainer(useContract);
+const Contracts = createContainer(useContracts);
 
-export default Contract;
+export default Contracts;
--- a/containers/DvmState.ts
+++ b/containers/DvmState.ts
@@ -25,10 +25,7 @@ const initState = {
 
 const useContractState = () => {
   const { block$ } = Connection.useContainer();
-  const {
-    votingContract: dvm,
-    storeContract: store,
-  } = DvmContract.useContainer();
+  const { votingContract, storeContract } = DvmContract.useContainer();
   const { address: collAddress } = Collateral.useContainer();
   const { empState } = EmpState.useContainer();
   const { priceIdentifier, expirationTimestamp } = empState;
@@ -39,19 +36,19 @@ const useContractState = () => {
   // get state from EMP
   const queryState = async () => {
     if (
-      dvm !== null &&
-      store !== null &&
+      votingContract !== null &&
+      storeContract !== null &&
       collAddress !== null &&
       priceIdentifier !== null &&
       expirationTimestamp !== null
     ) {
       const res = await Promise.all([
-        dvm.hasPrice(
+        votingContract.hasPrice(
           priceIdentifier.toString(),
           expirationTimestamp.toNumber(),
           { from: empAddress }
         ),
-        store.computeFinalFee(collAddress, { from: empAddress }),
+        storeContract.computeFinalFee(collAddress, { from: empAddress }),
       ]);
 
       const hasPrice = res[0] as boolean;
@@ -61,7 +58,7 @@ const useContractState = () => {
       if (hasPrice) {
         try {
           const postResolutionRes = await Promise.all([
-            dvm.getPrice(
+            storeContract.getPrice(
               priceIdentifier.toString(),
               expirationTimestamp.toNumber(),
               { from: empAddress }
@@ -86,15 +83,15 @@ const useContractState = () => {
   // get state on setting of contract
   useEffect(() => {
     queryState();
-  }, [dvm, priceIdentifier, expirationTimestamp]);
+  }, [storeContract, priceIdentifier, expirationTimestamp]);
 
   // get state on each block
   useEffect(() => {
-    if (block$ && dvm) {
+    if (block$ && storeContract) {
       const sub = block$.subscribe(() => queryState());
       return () => sub.unsubscribe();
     }
-  }, [block$, dvm]);
+  }, [block$, storeContract]);
 
   return { dvmState: state };
 };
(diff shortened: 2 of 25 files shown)


containers/DvmContracts.ts
```diff
@@ -0,0 +1,62 @@
+import { createContainer } from "unstated-next";
```
chrismaree, line 1: This whole file is shown as a diff because I re-named it. I chose to add the store state into the `DVMContracts.ts` rather than creating a separate store contract state container.

The main addition is that the `Promise.all` returns the `Store` address in addition to the `Oracle` address.

containers/DvmState.ts
```diff
@@ -12,24 +14,19 @@ const { formatUnits: fromWei } = utils;
...
+  finalFee: number | null;
 }
 
 const initState = {
   hasEmpPrice: null,
   resolvedPrice: null,
+  finalFee: null,
 };
 
-// Taken from Voting.sol implementation.
```
chrismaree, line 22: this was left over from when we were using theGraph to do some of these queries. I removed it as it's not needed any more.

containers/DvmState.ts
```diff
@@ -39,25 +36,29 @@ const useContractState = () => {
...
-      dvm !== null &&
+      votingContract !== null &&
+      storeContract !== null &&
+      collAddress !== null &&
       priceIdentifier !== null &&
       expirationTimestamp !== null
     ) {
       const res = await Promise.all([
-        dvm.hasPrice(
+        votingContract.hasPrice(
```
chrismaree, line 46: rather than calling it `dvm` I thought it cleaner to refer to the `votingContract` and `storeContract` as this now stores two contracts.

features/contract-state/GeneralInfo.tsx
```diff
@@ -124,14 +124,18 @@ const GeneralInfo = () => {
...
         <Status>
-          <Label>Is Expired: </Label>
-          <Link
-            href={DOCS_MAP.EXPIRY_SETTLEMENT}
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            <strong>{isExpired}</strong>
-          </Link>
+          <Label>
```
chrismaree, line 127: change made to be consistent. looks like this:
image

features/contract-state/DisputeParams.tsx
```diff
@@ -25,20 +27,38 @@ const fromWei = ethers.utils.formatUnits;
...
+          <Link
+            href={DOCS_MAP.FINAL_FEE}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            Docs
+          </Link>
+          ){`: `}
+        </Label>
+        {finalFee ? finalFee : "N/A"} {collSymbol}
```
chrismaree, line 60: note this component does not use the `renderComponent` notation. There are a few other places through the code that don't and I didn't want to muddle this PR too much with changing this so I used the same pattern as the rest of this file.

containers/DvmContracts.ts
nicholaspai: Makes sense to group Store and Voting into the same file

containers/DvmState.ts
```diff
@@ -39,25 +36,29 @@ const useContractState = () => {
...
           priceIdentifier.toString(),
           expirationTimestamp.toNumber(),
           { from: empAddress }
         ),
+        storeContract.computeFinalFee(collAddress, { from: empAddress }),
```
nicholaspai, line 51: nit: Does this have to be called from `empAddress`to succeed?

containers/DvmState.ts
```diff
@@ -39,25 +36,29 @@ const useContractState = () => {
...
       ]);
 
       const hasPrice = res[0] as boolean;
+      const finalFee = parseFloat(fromWei(res[1][0].toString()));
```
nicholaspai, line 55: Does this still work if the collateral does not have a final fee in the store?

nicholaspai: i.e. does it return 0

containers/DvmState.ts
```diff
@@ -81,15 +83,15 @@ const useContractState = () => {
   // get state on setting of contract
   useEffect(() => {
     queryState();
-  }, [dvm, priceIdentifier, expirationTimestamp]);
+  }, [votingContract, priceIdentifier, expirationTimestamp]);
```
nicholaspai, line 86: You should add `storeContract` to this array so that `useEffect()` updates if the store changes

containers/DvmState.ts
```diff
@@ -81,15 +83,15 @@ const useContractState = () => {
...
 
   // get state on each block
   useEffect(() => {
-    if (block$ && dvm) {
+    if (block$ && votingContract) {
```
nicholaspai, line 90: Also maybe add `storeContract` here for consistency

nicholaspai: Actually, I'm not sure you even need `votingContract` here, maybe you only need `block$`

features/contract-state/DisputeParams.tsx
nicholaspai: nits:
- This check should be `finalFee !== null`
- don't you also have to check if `collSymbol` is null?

containers/DvmState.ts
```diff
@@ -39,25 +36,29 @@ const useContractState = () => {
...
   const queryState = async () => {
     if (
-      dvm !== null &&
+      votingContract !== null &&
+      storeContract !== null &&
+      collAddress !== null &&
       priceIdentifier !== null &&
       expirationTimestamp !== null
     ) {
       const res = await Promise.all([
```
mrice32, line 45: Small stylistic comment (I know this code pre-dated this PR). Would this be easier to read if it were:
```
const [hasPriceResult, finalFeeResult] = Promise.all...
```

containers/DvmContracts.ts
```diff
@@ -0,0 +1,62 @@
...
+        uma.finder.abi,
+        provider
+      );
+      const res = await Promise.all([
+        finder.getImplementationAddress(utf8ToHex("Oracle")),
+        finder.getImplementationAddress(utf8ToHex("Store")),
+      ]);
+
+      const votingAddress = res[0];
+      const storeAddress = res[1];
```
mrice32, line 35: nit might be a little simpler:
```js
      const [votingAddress, storeAddress] = await Promise.all([
        finder.getImplementationAddress(utf8ToHex("Oracle")),
        finder.getImplementationAddress(utf8ToHex("Store")),
      ]);

containers/DvmState.ts
mrice32: Yep, I think it returns 0 in that case: https://github.com/UMAprotocol/protocol/blob/acf99518ed0f42756814989331a8ed592ab99ebe/core/contracts/oracle/implementation/Store.sol#L132-L134

containers/DvmState.ts
mrice32: Agreed.

features/contract-state/DisputeParams.tsx
chrismaree: done. will need to apply the same cleaning process on a bunch of other places through the dapp.

containers/DvmState.ts
chrismaree: nope, does not. good catch. have removed

containers/DvmState.ts
chrismaree: it should,yes.

containers/DvmState.ts
chrismaree: cool. I changed it to. PTAL

containers/DvmState.ts
chrismaree: done. PTAL

containers/DvmState.ts
chrismaree: ye, can change it to that but perhaps in a separate cleaning PR. I have created an issue to keep track of this point #106

chrismaree: I went ahead and added it to this PR anyways as it's pretty small

containers/DvmContracts.ts
chrismaree: implemented

chrismaree:
@mrice32 I've implemented your nits. PTAL.

PR #108, opened 2020-08-03
Display live liquidations for current sponsor's position

### LINKED ISSUE #105: Display liquidation data for positions
right now there is no way to see if a position's `health` is it being liquidated? pending transfer? this information should be shown.

nicholaspai:
Currently only displays live liquidations for sponsors as there are several cases I didn't want to have to catch yet such as when a liquidation is expired and its rewards have been withdrawn. We would no longer be able to query liquidation metadata via contract calls, and would instead need to query past events. For now, I've kept the implementation simple and only parse `getLiquidations()` return data for liquidations that are still live.

Resolves #105


nicholaspai:
User's Liquidation panel appears if they have a live liquidation:
Screen Shot 2020-08-03 at 18 13 53

User gets an alert on home page to check out "Manage Positions" and dive deeper into live liquidations:
Screen Shot 2020-08-03 at 18 13 44

features/contract-state/ContractState.tsx
```diff
@@ -1,11 +1,29 @@
-import { Box, Grid } from "@material-ui/core";
+import { Box, Grid, Typography } from "@material-ui/core";
+import styled from "styled-components";
 import GeneralInfo from "./GeneralInfo";
 import DisputeParams from "./DisputeParams";
 import Totals from "./Totals";
 
+import Position from "../../containers/Position";
+
+const Important = styled(Typography)`
```
chrismaree, line 9: general meta-comment, unrelated to this PR: I feel that we should pull all these styling blocks up into a general styling file. this `Important` styling, for example, is used in a few other places.

nicholaspai: Agreed, I'll create an issue for this as I don't want to drasticallyt increase diff on this PR

PR #109, opened 2020-08-03
Add Liquidation Table under AllSponsors table

### LINKED ISSUE #105: Display liquidation data for positions

nicholaspai:
Also refactors the positions table into its own component, AllSponsors. AllPositions now displays AllLiquidations beneath AllSponsors.

Follow up to #108  ~and draft mode until #108 is merged.~

Resolves #105


nicholaspai:
Main liquidation table screenshot:
Screen Shot 2020-08-03 at 19 05 59

More data if scroll to the right:
Screen Shot 2020-08-03 at 19 06 04

If liquidation has been disputed, links to disputer address:
Screen Shot 2020-08-03 at 19 06 14

features/manage-position/YourLiquidations.tsx
```diff
@@ -56,7 +56,7 @@ const YourLiquidations = () => {
     liquidations.map((liq: LiquidationState) => {
       let maxDisputablePrice =
         liq.tokensOutstanding > 0 && collReqFromWei > 0
-          ? liq.liquidatedCollateral / (liq.tokensOutstanding / collReqFromWei)
+          ? liq.liquidatedCollateral / (liq.tokensOutstanding * collReqFromWei)
```
nicholaspai, line 59: Outside scope of this PR but this was an easy fix to a typo from my latest YourLiquidations PR.

chrismaree:
This is pretty awesome, great work!
drive-by comments on styling:
1) I dont think this table should always show up; in fact I think it should be hitten with a button that lets you swap between the all positions table and the liquidations table. This could be a switch at the top or a button at the bottom of the table that says `show liquidations`
2) From your screenshot, it says `PreDispute` for liquidations that have gone past the liquidation liveness. This is pretty confusing and should perhaps show something different like `liquidation succeeded`. This'll be a combination of the liquidation state, liveness and liquidation time.
3) I don't think we should show `liquidation receipt`, `liquidator` or `disputer` in the table. It makes it hard to read. ideally, you don't need to scroll sideways on the table. You can leave it in for now but I defs think it should be re-factored into a modal that gives you "extra" info if you click on the liquidation. That'll be similar to the click to liquidate modal when that PR is finished up.

nicholaspai:
> I dont think this table should always show up; in fact I think it should be hitten with a button that lets you swap between the all positions table and the liquidations table. This could be a switch at the top or a button at the bottom of the table that says show liquidations

I added a checkbox "Show Liquidations" that will render the liquidations table and is NOT checked by default (i.e. liquidation is hidden). It will show the liquidations table below the All sponsors table, which is showed by default, but I don't think we should only show one--instead give user to render both. WDYT? Play around with preview pls

> From your screenshot, it says PreDispute for liquidations that have gone past the liquidation liveness. This is pretty confusing and should perhaps show something different like liquidation succeeded. This'll be a combination of the liquidation state, liveness and liquidation time.

I agreed and added a translation switch, preview:
Screen Shot 2020-08-05 at 08 01 17

> I don't think we should show liquidation receipt, liquidator or disputer in the table. It makes it hard to read. ideally, you don't need to scroll sideways on the table. You can leave it in for now but I defs think it should be re-factored into a modal that gives you "extra" info if you click on the liquidation. That'll be similar to the click to liquidate modal when that PR is finished up.

I think we should include the receipt/tx hash if anything, but we should probably leave out `liquidator`, `disputer`, `liquidationId`. What else?

apollo/uma/queries.ts
```diff
@@ -11,6 +11,27 @@ export const ACTIVE_POSITIONS = gql`
...
+        }
+        disputer {
+          address
+        }
+        tokensLiquidated
+        lockedCollateral
+        liquidatedCollateral
+        status
+        events {
+          __typename
```
mrice32, line 30: Just curious -- what other types of events are in the liquidation bucket?

containers/EmpSponsors.ts
```diff
@@ -116,7 +136,43 @@ const useEmpSponsors = () => {
...
+                    parseFloat(liquidation.tokensLiquidated)
+                  : 0;
+              const maxDisputablePrice =
+                parseFloat(liquidation.tokensLiquidated) > 0 &&
+                collReqFromWei > 0
+                  ? parseFloat(liquidation.liquidatedCollateral) /
+                    (parseFloat(liquidation.tokensLiquidated) * collReqFromWei)
+                  : 0;
+              // There should always be a LiquidationCreatedEvent associated with each liquidation object, but if not then
+              // we will just ignore this strange edge case.
```
mrice32, line 157: nit: should this comment be at the top of this if statement?

features/all-positions/AllLiquidations.tsx
```diff
@@ -0,0 +1,330 @@
...
+  // Set max page depending on # of liqs
+  useEffect(() => {
+    setMaxPage(1);
+    setPage(1);
+
+    if (liquidations) {
+      let extraPages = 1;
+      if (Object.keys(liquidations).length % ITEMS_PER_PAGE === 0) {
+        extraPages = 0;
+      }
```
mrice32, line 107: tiny nit -- might be a little simpler:
```
const extraPages = Object.keys(liquidations).length % ITEMS_PER_PAGE === 0 ? 0 : 1;
```

features/all-positions/AllLiquidations.tsx
```diff
@@ -0,0 +1,330 @@
...
+      setMaxPage(
+        Math.floor(Object.keys(liquidations).length / ITEMS_PER_PAGE) +
+          extraPages
+      );
+      // This will set maxPage to 0 if there are no liqs.
```
mrice32, line 112: nit: won't it set it to 1 (what it's set to at the top of this function before the if)?

features/all-positions/AllLiquidations.tsx
```diff
@@ -0,0 +1,330 @@
...
+    // First filters out liq. data missing field values,
+    // then sorts the positions according to selected sort column,
+    // and finally slices the array based on pagination selection.
+    const reformattedLiquidationData = Object.keys(liquidations)
+      .filter((sponsor: string) => {
+        return (
+          liquidations[sponsor]?.maxDisputablePrice &&
+          liquidations[sponsor]?.tokensLiquidated &&
+          liquidations[sponsor]?.lockedCollateral &&
+          liquidations[sponsor]?.liquidationTimestamp
```
mrice32, line 183: Quick question: it's desired to filter these out if any of them are 0 in addition to being null, right?

features/all-positions/AllLiquidations.tsx
```diff
@@ -0,0 +1,330 @@
...
+    };
+
+    const invertDisputablePrice = isPricefeedInvertedFromTokenSymbol(
+      tokenSymbol
+    );
+    const getDisputablePrice = (x: string) => {
+      return invertDisputablePrice && parseFloat(x) !== 0
+        ? 1 / Number(x)
+        : Number(x);
+    };
```
mrice32, line 146: Slightly confused -- why would you need to invert the price here? Isn't the input always tokens/collateral since you're pulling the amounts directly from the liquidation? Or am I missing a step?

apollo/uma/queries.ts
nicholaspai: From the GraphQL schema: 

Screen Shot 2020-08-05 at 17 09 45

containers/EmpSponsors.ts
nicholaspai: yeah definitely

features/all-positions/AllLiquidations.tsx
nicholaspai: Yup agree, will make this change for `AllSponsors` as well

features/all-positions/AllLiquidations.tsx
nicholaspai: Yeah agreed I removed this comment

features/all-positions/AllLiquidations.tsx
nicholaspai: Yeah I suppose thats right. We can filter them out in the `EMPSponsors` container

features/all-positions/AllLiquidations.tsx
nicholaspai: The input is: 

```
parseFloat(liquidation.liquidatedCollateral) /
                    (parseFloat(liquidation.tokensLiquidated) * collReqFromWei)
```

Which sets the max price for the liquidation timestamp for it to be disputable (i.e. if the price turns out to be below this, then you can dispute the price).

For USDETH this might be 0.0028 ($359 = 1 ETH), but this dApp performs inversions on pricefeeds that are more "intuitively" read inverted such as USDETH --> ETHUSD. So, the "max disputable price" is 0.0028 but this implies a "min disputable price" of $359 (i.e. if the price is above 359 then the liquidation is disputable).

And we want to display $359 to the user instead of 0.0028.

Does this make sense?

features/all-positions/AllSponsors.tsx
```diff
@@ -0,0 +1,250 @@
+import {
```
chrismaree, line 1: splitting these is good. will fit well in the other PR with click to liquidate functions.

apollo/uma/queries.ts
```diff
@@ -11,6 +11,27 @@ export const ACTIVE_POSITIONS = gql`
           id
         }
       }
+      liquidations {
```
chrismaree, line 14: so query all this information, every time. do we have any understanding of how much data this query is now transferring? not really an issue with 200 or 300 sponsors but might become a bit of an issue if we scale up more and need to improve the pagination of the table to query more data as you move between pages.

containers/EmpSponsors.ts
```diff
@@ -107,16 +127,64 @@ const useEmpSponsors = () => {
...
+            }
+          });
+
+          empData.liquidations.forEach((liquidation: LiquidationQuery) => {
+            const sponsor = utils.getAddress(liquidation.sponsor.id);
+            const liquidationCreatedEvent = liquidation.events.find(
+              (e) => e.__typename === "LiquidationCreatedEvent"
+            );
+
+            // There should always be a LiquidationCreatedEvent associated with each liquidation object, but if not then
```
chrismaree, line 150: if it is always the case then why include the if statement? how would there be a liquidation without the event?

apollo/uma/queries.ts
nicholaspai: I think it will do the same as the `positions` query which is to query the first 100 by default. I didn't worry about this for now because I don't expect there to be many liquidations relative to sponsor positions. WDYT?

I will explicitly set this to `first: 1000` for now just in case

containers/EmpSponsors.ts
nicholaspai: It'd be an indexing error but we should check just incase. If there for whatever reason is no `LiquidationCreatedEvent` then we can't trust any of the other data

PR #115, opened 2020-08-07
EMPSponsors Container: use sponsor+liqID as key instead of sponsor to display multiple liquidations for sponsors

nicholaspai:
Look at the row for sponsor beginning 0x6B for the difference:

Current site only shows one liquidation for a sponsor:
Screen Shot 2020-08-07 at 08 54 16

This fix slightly modifies the key used to store liquidation data to capture multiple liqs per sponsor:
Screen Shot 2020-08-07 at 08 53 58

### DIFF
--- a/containers/EmpSponsors.ts
+++ b/containers/EmpSponsors.ts
@@ -23,6 +23,10 @@ interface SponsorMap {
   [sponsor: string]: SponsorPositionState;
 }
 
+interface LiquidationMap {
+  [sponsorPlusId: string]: SponsorPositionState;
+}
+
 // Interfaces for GraphQl queries.
 interface PositionQuery {
   sponsor: { id: string };
@@ -31,6 +35,7 @@ interface PositionQuery {
 }
 
 interface LiquidationQuery {
+  id: string;
   sponsor: { id: string };
   liquidationId: string;
   liquidator: { address: string };
@@ -106,7 +111,7 @@ const useEmpSponsors = () => {
 
         if (empData) {
           let newPositions: SponsorMap = {};
-          let newLiquidations: SponsorMap = {};
+          let newLiquidations: LiquidationMap = {};
 
           const collReqFromWei = parseFloat(
             fromWei(collateralRequirement, collDecs)
@@ -142,7 +147,6 @@ const useEmpSponsors = () => {
           });
 
           empData.liquidations.forEach((liquidation: LiquidationQuery) => {
-            const sponsor = utils.getAddress(liquidation.sponsor.id);
             const liquidationCreatedEvent = liquidation.events.find(
               (e) => e.__typename === "LiquidationCreatedEvent"
             );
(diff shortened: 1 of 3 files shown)


containers/EmpSponsors.ts
```diff
@@ -23,6 +23,10 @@ interface SponsorMap {
   [sponsor: string]: SponsorPositionState;
 }
 
+interface LiquidationMap {
+  [sponsorPlusId: string]: SponsorPositionState;
```
mrice32, line 27: Minor naming nit: I'm getting a little confused where we're using the term `id` to refer to `1` vs `0x1e17a75616cd74f5846b1b71622aa8e10ea26cc0-1`. Here, the convention seems to be:
```
// Id = sponsor liquidation id
Id: 1

// Sponsor + id = 0x1e17a75616cd74f5846b1b71622aa8e10ea26cc0-1
sponsorPlusId: 0x1e17a75616cd74f5846b1b71622aa8e10ea26cc0-1
```

However, in the code below, it seems that you're using the following convention:
```
// Id = sponsor address + sponsor liquidation id
Id: 0x1e17a75616cd74f5846b1b71622aa8e10ea26cc0-1
```

Might make sense to just come up with a consistent term for each type of id for clarity.

nicholaspai: Yeah the issue here is that the subgraph uses `id` to reference `sponsorAddress+ID`. Is this the line that's specifically confusing you, where I key a liquidation object by its returned `liquidation.id`?  If so, then if I set `liquidation.id` to a more clear variable like `const sponsorAddressPlusId` does it make it clearer?

```
                const sponsorAddressPlusId = liquidation.id
                newLiquidations[sponsorAddressPlusId] = {
                  sponsor: utils.getAddress(liquidation.sponsor.id),
                  liquidator: liquidation.liquidator?.address,
                  disputer: liquidation.disputer?.address,
                  liquidationId: liquidation.liquidationId,
                  liquidatedCR: liquidatedCR.toString(),
                  maxDisputablePrice: maxDisputablePrice.toString(),
                  tokensLiquidated: liquidation.tokensLiquidated,
                  lockedCollateral: liquidation.lockedCollateral,
                  status: liquidation.status,
                  liquidationTimestamp: liquidationCreatedEvent.timestamp,
                  liquidationReceipt: liquidationCreatedEvent.tx_hash,
                };

```

features/all-positions/AllLiquidations.tsx
```diff
@@ -172,19 +172,19 @@ const AllLiquidations = () => {
...
+      .filter((sponsorAddressPlusId: string) => {
         return (
-          liquidations[sponsor]?.maxDisputablePrice &&
-          liquidations[sponsor]?.tokensLiquidated &&
-          liquidations[sponsor]?.lockedCollateral &&
-          liquidations[sponsor]?.liquidationTimestamp
+          liquidations[sponsorAddressPlusId]?.maxDisputablePrice &&
+          liquidations[sponsorAddressPlusId]?.tokensLiquidated &&
+          liquidations[sponsorAddressPlusId]?.lockedCollateral &&
+          liquidations[sponsorAddressPlusId]?.liquidationTimestamp
```
mrice32, line 180: Thanks, this is easier to understand.

nit: just as a shortener, maybe we could do `sponsorPlusId` as you did above?

PR #117, opened 2020-08-11
"Max Buttons" incorrectly calculate max amounts to Redeem and Unwrap-Weth

### LINKED ISSUE #116: Redeem should set amounts > num Tokens outstanding to num tokens outstanding
Redeem fails if amount to redeem > # tokens outstanding, the "MAX" button should be set to this limit

nicholaspai:
There are two problems here:
- The "Max" button in the Redeem feature sets the max amount to redeem equal to the user's balance, even if they have more tokens than # tokens outstanding in their position. This needs to be set to `Math.min(tokenBalance, tokensInPosition)`
- The amount set to the max is equal to the "intended max" amount casted to a number, and JS casting loses precision. Therefore, the amount set to the max should be equal to the string. This affects both Redeem and UnwrapWETH
Fixes #116

### DIFF
--- a/containers/Position.ts
+++ b/containers/Position.ts
@@ -9,16 +9,16 @@ import Collateral from "./Collateral";
 import Token from "./Token";
 
 const fromWei = ethers.utils.formatUnits;
-const weiToNum = (x: BigNumberish, u = 18) => parseFloat(fromWei(x, u));
+const weiToNum = (x: BigNumberish, u = 18) => fromWei(x, u);
 
 export interface LiquidationState {
   liquidator: string;
-  liquidatedCollateral: number;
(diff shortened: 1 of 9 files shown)


nicholaspai:
> The types we are dealing with here are pretty janky. it's such a pain to keep having to convert and _name_ the type within the variable. I wish there was a cleaner pattern for this.

Maybe a better solution is to export both a `string` version and a `number` version. This way if we don't want to lose precision we can use `string`, else `number`, wdyt

nicholaspai:
@chrismaree I'd like to tackle the type casting in a future PR, but want to make this bug fix for now: https://github.com/UMAprotocol/emp-tools/issues/118

PR #120, opened 2020-08-13
Add kovan contracts

### LINKED ISSUE #86: Support Kovan contracts
We have deployed our contracts + many EMP's to the Kovan network. We should be able to switch networks on our web3 provider when using the dApp seamlessly

nicholaspai:
- Switch seamlessly between Kovan and Mainnet contracts for WETH and EMP contracts
- Refuse to render `Yield` page when on Kovan because most of the balancer pool data here is irrelevant to Kovan
- Add support for Kovan subgraph URL

Question: 
- Is it possible to specify via `onboard.js` which networks we support? For example, I'd like this warning to show only if the user's network is not mainnet or kovan:
Screen Shot 2020-08-13 at 15 27 34

Other:
- Fixes bug in "Click To Dispute" feature where liquidation ID is incorrectly passed into `dispute()` method

Fixes #86

### DIFF
--- a/containers/EmpSponsors.ts
+++ b/containers/EmpSponsors.ts
@@ -75,10 +75,7 @@ const useEmpSponsors = () => {
   // We set the poll interval to a very slow 5 seconds for now since the position states
   // are not expected to change much.
   // Source: https://www.apollographql.com/docs/react/data/queries/#polling
-  let clientName = "UMA";
-  if (network !== null) {
-    clientName += network.chainId.toString();
-  }
+  const clientName = `UMA${network?.chainId.toString()}`;
   const { loading, error, data } = useQuery(EMP_DATA, {
     context: { clientName: clientName },
     pollInterval: 5000,
--- a/containers/EmpSponsors.ts
+++ b/containers/EmpSponsors.ts
@@ -3,6 +3,7 @@ import { useState, useEffect } from "react";
 import { utils } from "ethers";
 const fromWei = utils.formatUnits;
 
+import Connection from "./Connection";
 import EmpContract from "./EmpContract";
 import EmpState from "./EmpState";
 import Collateral from "./Collateral";
@@ -13,6 +14,7 @@ import { isPricefeedInvertedFromTokenSymbol } from "../utils/getOffchainPrice";
 
 import { useQuery } from "@apollo/client";
 import { EMP_DATA } from "../apollo/uma/queries";
+import { client } from "../apollo/client";
 
 // Interfaces for dApp state storage.
 interface SponsorPositionState {
@@ -61,6 +63,7 @@ interface FinancialContractQuery {
 }
 
 const useEmpSponsors = () => {
+  const { network } = Connection.useContainer();
   const { contract: emp } = EmpContract.useContainer();
   const { empState } = EmpState.useContainer();
   const { collateralRequirement } = empState;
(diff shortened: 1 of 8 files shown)


features/all-positions/LiquidationActionDialog.tsx
```diff
@@ -212,7 +212,7 @@ const LiquidationActionDialog = (props: DialogProps) => {
         try {
           if (needCollateralAllowance()) await setMaxCollateralAllowance();
           const tx = await emp.dispute(
-            liquidatedPosition.sponsorPlusId,
+            liquidatedPosition.liquidationId,
```
nicholaspai, line 215: @chrismaree This was a bug in `dispute()` method, ID should be passed not the sponsorPlusId

chrismaree: good catch, thanks for fixing!

features/yield/Yield.tsx
```diff
@@ -22,63 +26,74 @@ const Yield = () => {
...
-          more about yUSD see the UMA Medium post{" "}
-          <a
-            href="https://medium.com/uma-project/the-yield-dollar-on-uma-3a492e79069f"
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            here
-          </a>
-          .
+          <i>Please first connect and set your network to Mainnet.</i>
```
chrismaree, line 34: this is great. good change here for this case.

containers/EmpSponsors.ts
```diff
@@ -72,8 +75,12 @@ const useEmpSponsors = () => {
   // We set the poll interval to a very slow 5 seconds for now since the position states
   // are not expected to change much.
   // Source: https://www.apollographql.com/docs/react/data/queries/#polling
+  let clientName = "UMA";
+  if (network !== null) {
+    clientName += network.chainId.toString();
```
chrismaree, line 80: nit: could you not make this a const and ditch the line 78 with:
```
const clientName = `UMA${network.chainId.toString()}
```

containers/EmpSponsors.ts
```diff
@@ -72,8 +75,12 @@ const useEmpSponsors = () => {
...
+  }
   const { loading, error, data } = useQuery(EMP_DATA, {
-    context: { clientName: "UMA" },
+    context: { clientName: clientName },
```
chrismaree, line 83: nit: if the key and value is the same then do you need to include both? could this not simply be:
```
{ clinetName }
```

chrismaree:
This looks pretty good! For your question what you should do is link into [this](https://github.com/UMAprotocol/emp-tools/blob/fc24481d88c96dd65549fdc4f0e8f684d5467814/containers/Connection.ts#L34) function in `Connection.ts` and add the check if the network ID is part of the known list. Something like: `validNetworks.includes(networkId)`

containers/EmpSponsors.ts
nicholaspai: Yeah this is doable because if the `clientName !== UMA42` it will default to the mainnet URL. I think node.js calculates `${undefined}` as `undefined` though so it would just return `UMAundefined` which is OK.

containers/EmpSponsors.ts
nicholaspai: Yeah but I like explicitly defining it so that the dev knows that `clientName` is the name of the key. I'll change `clientName` to `subgraphToQuery`

nicholaspai:
> This looks pretty good! For your question what you should do is link into [this](https://github.com/UMAprotocol/emp-tools/blob/fc24481d88c96dd65549fdc4f0e8f684d5467814/containers/Connection.ts#L34) function in `Connection.ts` and add the check if the network ID is part of the known list. Something like: `validNetworks.includes(networkId)`

I implemented a very simple `alert()` message if user switches away from invalid networks, do you think there's something more complex we should do for now? I'm thinking that this is a relatively rare edge case and we can just be direct and obtrusive with this warning as the dApp will crash otherwise

Issue #122, opened 2020-08-18
Liquidation action modal (and liquidation table) should consider withdrawal requests in liquidation price calculation


chrismaree:
Right now a pending withdrawal is not factored into the position CR calculation in the modal nor the table. For example this sponsor:
![image](https://user-images.githubusercontent.com/12886084/90500309-fdcb2580-e14a-11ea-8144-5d8d7fbcbfb7.png)

Has a pending withdrawal that drops their CR. however, the table nor the modal reflects this in the CR calculation.

nicholaspai:
This should also be fixed in the YourPosition tab

PR #123, opened 2020-08-18
Reflect withdrawal request in CR and backing collateral for token debt

nicholaspai:
- Display requested withdrawal amount in YourPosition
- Calculate CR using req. withdrawal amount
- Display backing collateral (locked collateral - req withdrawal amount) in YourPosition

Part of #122 

Preview:
Screen Shot 2020-08-18 at 13 17 47

### DIFF
--- a/features/manage-position/YourPosition.tsx
+++ b/features/manage-position/YourPosition.tsx
@@ -39,6 +39,7 @@ const YourPosition = () => {
   const {
     tokens: tokenString,
     collateral: collString,
+    backingCollateral: backingCollString,
     cRatio,
     withdrawAmt: withdrawAmtString,
     pendingWithdraw,
@@ -54,6 +55,7 @@ const YourPosition = () => {
   if (
     tokenString !== null &&
     collString !== null &&
+    backingCollString !== null &&
     cRatio !== null &&
     gcr !== null &&
     collSymbol !== null &&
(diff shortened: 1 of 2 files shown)


features/manage-position/YourPosition.tsx
```diff
@@ -126,7 +131,11 @@ const YourPosition = () => {
               {`${_collateral} ${_collSymbol}`}
             </Status>
             <Status>
-              <Label>Tokens outstanding: </Label>
+              <Label>Collateral backing debt: </Label>
```
chrismaree, line 134: I thought about adding casework here to show the old text if you don't have a withdrawal pending and show this text if you do. However, I actually think it's cleaner like you've got it here because it becomes very clear exactly what your risk exposure is.

features/manage-position/YourPosition.tsx
```diff
@@ -72,9 +74,10 @@ const YourPosition = () => {
     const collReqFromWei = parseFloat(fromWei(collReq, collDec));
     const tokens = Number(tokenString);
     const collateral = Number(collString);
+    const backingCollateral = Number(backingCollString);
     const withdrawAmt = Number(withdrawAmtString);
     const liquidationPrice = getLiquidationPrice(
-      collateral,
+      collateral - withdrawAmt,
```
chrismaree, line 80: nice and simple.

features/manage-position/YourPosition.tsx
nicholaspai: yeah might as well show everything

PR #125, opened 2020-08-19
Creates: Autofill collateral to minimum required by GCR constraint

nicholaspai:
The goal is to encourage sponsor capital efficiency and create tokens using minimum "safe" amount of collateral (as required by GCR)
- Move "Tokens" text field left of "Collateral" text field since it is more intuitive to think of how much debt one wants to take out
- Auto-fill amount of collateral backing desired amount of token debt to create to the minimum allowed by the GCR
- Add a "min" button to set collateral amount to min

Open question:
- Is it more intuitive to auto fill the "min" backing collateral or the "max" token debt? I.e., instead of a "min" button in the collateral text field, should we have a "max" button in the token text field?

Preview on Kovan:
Screen Shot 2020-08-19 at 11 45 30

### DIFF
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -1,4 +1,4 @@
-import { ethers, utils } from "ethers";
+import { utils } from "ethers";
 import styled from "styled-components";
 import {
   Box,
@@ -7,10 +7,11 @@ import {
   Typography,
   Grid,
   Tooltip,
+  InputAdornment,
 } from "@material-ui/core";
 
 import EmpContract from "../../containers/EmpContract";
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import Collateral from "../../containers/Collateral";
 import Token from "../../containers/Token";
 import EmpState from "../../containers/EmpState";
@@ -32,6 +33,10 @@ const Important = styled(Typography)`
 const Link = styled.a`
   color: white;
   font-size: 14px;
+`;
+
+const MinLink = styled.div`
+  text-decoration-line: underline;
 `;
 
 const {
@@ -159,6 +164,17 @@ const Create = () => {
       }
     };
 
+    const setBackingCollateralToMin = () => {
+      // By default set amount of collateral to the minimum required by the GCR constraint. This
+      // default is intended to encourage users to maximize their capital efficiency.
+      const minBackingCollateral = gcr * Number(tokens);
+      setCollateral(minBackingCollateral.toString());
+    };
+
+    useEffect(() => {
+      setBackingCollateralToMin();
+    }, [gcr, tokens, minSponsorTokensFromWei]);
+
     if (hasPendingWithdraw) {
       return (
         <Box py={2}>
@@ -216,24 +232,6 @@ const Create = () => {
                 fullWidth
                 type="number"
                 variant="outlined"
-                label={`Collateral (${collSymbol})`}
-                inputProps={{ min: "0", max: balance }}
-                value={collateral}
-                error={balanceBelowCollateralToDeposit}
-                helperText={
-                  balanceBelowCollateralToDeposit &&
-                  `${collSymbol} balance is too low`
-                }
-                onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
-                  setCollateral(e.target.value)
-                }
-              />
-            </Grid>
-            <Grid item md={4} sm={6} xs={12}>
-              <TextField
-                fullWidth
-                type="number"
-                variant="outlined"
                 label={`Tokens (${tokenSymbol})`}
                 inputProps={{ min: "0" }}
                 value={tokens}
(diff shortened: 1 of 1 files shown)


features/manage-position/Create.tsx
```diff
@@ -159,6 +164,17 @@ const Create = () => {
...
+    const setBackingCollateralToMin = () => {
+      // By default set amount of collateral to the minimum required by the GCR constraint. This
+      // default is intended to encourage users to maximize their capital efficiency.
+      const minBackingCollateral = gcr * Number(tokens);
+      setCollateral(minBackingCollateral.toString());
+    };
+
+    useEffect(() => {
+      setBackingCollateralToMin();
+    }, [gcr, tokens, minSponsorTokensFromWei]);
```
mrice32, line 176: If this gets updated anytime the gcr changes, does that mean that if you're looking at this page as someone else is creating or redeeming a position, the collateral field might update unexpectedly?

There may not be a better way to do it, but just wanted to make sure I'm understanding this correctly.

nicholaspai: I believe that's right, we could basically remove `gcr` from this so that it just updates once on render but I see this as an edge case that shouldn't cause problems

features/manage-position/Create.tsx
```diff
@@ -159,6 +164,17 @@ const Create = () => {
       }
     };
 
+    const setBackingCollateralToMin = () => {
+      // By default set amount of collateral to the minimum required by the GCR constraint. This
+      // default is intended to encourage users to maximize their capital efficiency.
+      const minBackingCollateral = gcr * Number(tokens);
+      setCollateral(minBackingCollateral.toString());
```
chrismaree, line 171: should we consider rounding this to 4 decimal places? it looks a bit nasty in the UI IMHO for these numbers:
image

Otherwise, you can round in the button or input field.

nicholaspai: I rounded it to 4

features/manage-position/Create.tsx
```diff
@@ -73,6 +78,19 @@ const Create = () => {
   } = empState;
   const liquidationPriceWarningThreshold = 0.1;
 
+  useEffect(() => {
```
nicholaspai, line 81: Can't call `useEffect()` or other hooks inside of conditionals: https://medium.com/@jonchurch/how-to-fix-react-error-rendered-fewer-hooks-than-expected-e6a378985d3c 

Otherwise you can get a `Rendered fewer hooks than expected` error

features/manage-position/Create.tsx
```diff
@@ -73,6 +78,19 @@ const Create = () => {
...
+    if (gcr !== null) {
+      setBackingCollateralToMin(gcr, Number(tokens));
+    }
+  }, [gcr, tokens]);
+
+  const setBackingCollateralToMin = (_gcr: number, _tokens: number) => {
+    // By default set amount of collateral to the minimum required by the GCR constraint. This
+    // default is intended to encourage users to maximize their capital efficiency.
+    const minBackingCollateral = _gcr * _tokens;
+    setCollateral(minBackingCollateral.toFixed(4));
```
mrice32, line 91: nit: `toFixed()` rounds, right? So if it rounds down, will the pre-filled collateral not meet the GCR requirement?

features/manage-position/Create.tsx
mrice32: Yeah, I think this is fine, and I don't see an easy way to avoid it unless you _only_ fill it when the tokens field is modified.

features/manage-position/Create.tsx
nicholaspai: yeah good point, but there is no clean solution to "round up": https://helloacm.com/javascripts-tofixed-implementation-without-rounding/

What I can do is naively just add 0.00005 to the number before sending to `.toFixed()`

nicholaspai: Actually, the cleaner solution might be to NOT round in this function and just round in the JSX render element

features/manage-position/Create.tsx
```diff
@@ -73,6 +78,22 @@ const Create = () => {
...
+  }, [gcr, tokens]);
+
+  const setBackingCollateralToMin = (_gcr: number, _tokens: number) => {
+    // By default set amount of collateral to the minimum required by the GCR constraint. This
+    // default is intended to encourage users to maximize their capital efficiency.
+    const minBackingCollateral = _gcr * _tokens;
+    // We want to round down the number for better UI display, but we don't actually want the collateral
+    // amount to round down since we want the minimum amount of collateral to pass the GCR constraint. So,
+    // we'll add a tiny amount of collateral to avoid accidentally rounding too low.
+    setCollateral((minBackingCollateral + 0.00005).toFixed(4));
```
chrismaree, line 94: what are the implications of hard coding this if we have non-18 decimal collateral type? `0.00005` ether is not much but `0.00005` is like 50c. it's also probs fine but just though i'd mention it.

chrismaree (approved):
looks great! good defaulting is powerful!

PR #127, opened 2020-08-19
feat(AllPositions): liquidation and position dialog enhancements

### LINKED ISSUE #121: Liquidation action modal does not show the correct liquidation ID
See screenshot below. Modal is missing liquidation ID. 
image

### LINKED ISSUE #122: Liquidation action modal (and liquidation table) should consider withdrawal requests in liquidation price calculation

### LINKED ISSUE #126: Positions dialog should set reasonable default values for min/maxCollPerToken and liquidation deadline
These function params are not obvious and we should set them to default values

nicholaspai:
Position Dialog:
- CRatio and LiquidationPrice should account for requested withdrawal amount. Fixes #122 
- Should show backing collateral, not just locked collateral (accounts for withdrawal reqs)
- "Click to Liquidate" dialog should auto fill default values for `minCollPerToken` and `maxCollPerToken` as well as `liquidationDeadline`. Fixes #126 
- Preview: 
Screen Shot 2020-08-19 at 12 36 30

Liquidation Dialog: 
- Should show backing collateral, not just locked collateral (accounts for withdrawal reqs)
- (bug) Liquidation ID displayed correctly. Fixes #121 
- Preview: 
Screen Shot 2020-08-19 at 12 20 34

### DIFF
--- a/features/all-positions/PositionActionsDialog.tsx
+++ b/features/all-positions/PositionActionsDialog.tsx
@@ -1,7 +1,7 @@
 import styled from "styled-components";
 import { utils } from "ethers";
 const { formatUnits: fromWei, parseUnits: toWei } = utils;
-import { useState, MouseEvent } from "react";
+import { useState, MouseEvent, useEffect } from "react";
 
 import {
   Box,
@@ -283,6 +283,19 @@ const PositionActionsDialog = (props: DialogProps) => {
         setError(new Error("Please check that you are connected."));
       }
     };
+
+    useEffect(() => {
+      // Set min and max collPerToken params for the Liquidation dialog to
+      // approximately the current CRatio, +/- some error threshold.
+      const collPerToken =
+        Number(sponsorPosition.backingCollateral) /
+        Number(sponsorPosition.tokensOutstanding);
+      setMinCollPerToken((collPerToken * 0.95).toFixed(10));
+      setMaxCollPerToken((collPerToken * 1.05).toFixed(10));
+
+      // Set liquidation transaction deadline to a reasonable 30 mins to wait for it to be mined.
+      setDeadline((30 * 60).toString());
+    }, [sponsorPosition]);
 
     return (
       <Dialog open={props.isDialogShowing} onClose={props.handleClose}>
--- a/features/all-positions/PositionActionsDialog.tsx
+++ b/features/all-positions/PositionActionsDialog.tsx
@@ -15,6 +15,7 @@ import {
   InputLabel,
   FormControl,
   MenuItem,
+  Tooltip,
 } from "@material-ui/core";
 
 import ToggleButton from "@material-ui/lab/ToggleButton";
@@ -304,6 +305,17 @@ const PositionActionsDialog = (props: DialogProps) => {
                   {prettyBalance(Number(sponsorPosition.collateral))}
                 </Status>
                 <Status>
+                  <Tooltip
+                    title={
+                      "Equal to the position collateral minus any requested withdrawal amount"
+                    }
+                    placement="top"
+                  >
+                    <Label>Backing Collateral({collSymbol}): </Label>
+                  </Tooltip>
+                  {prettyBalance(Number(sponsorPosition.backingCollateral))}
+                </Status>
+                <Status>
                   <Label>Minted Synthetics({tokenSymbol}): </Label>
                   {prettyBalance(Number(sponsorPosition.tokensOutstanding))}
                 </Status>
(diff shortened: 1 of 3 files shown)


features/all-positions/PositionActionsDialog.tsx
```diff
@@ -283,6 +284,19 @@ const PositionActionsDialog = (props: DialogProps) => {
...
     };
 
+    useEffect(() => {
+      // Set min and max collPerToken params for the Liquidation dialog to
+      // approximately the current CRatio, +/- some error threshold.
+      const collPerToken =
+        Number(sponsorPosition.backingCollateral) /
+        Number(sponsorPosition.tokensOutstanding);
+      setMinCollPerToken((collPerToken * 0.95).toFixed(10));
+      setMaxCollPerToken((collPerToken * 1.05).toFixed(10));
```
mrice32, line 294: nit: I think these are strange defaults, but they might be _good enough_.

Intuitively, the min collateral should be the point at which you'll lose money on the liquidation (i.e. the total collateral in the position should be > 100% + some cushion) and the max collateral should be the point at which the liquidation becomes invalid (i.e. the total collateral - withdraw request should be < 125% - some cushion). Do you think it would be hard to compute those values? If so, it may not be worth it, but just wanted to mention it.

nicholaspai: This could be possible to calculate but currently we do not even block liquidations if they would be unprofitable. Perhaps we change this in a separate PR @chrismaree wdyt?

nicholaspai: How would you calculate the smart min and max's? I agree that those should be the thresholds

nicholaspai: Implemented actually pretty easy to calculate. See comments for details

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -103,6 +104,40 @@ const PositionActionsDialog = (props: DialogProps) => {
...
+      //   the amount of backing collateral in the position must be >= 100% of the token debt.
+      //   Therefore, the minimum `collPerToken` is simply the current price of 1 unit token debt,
+      //   which is the `latestPrice`.
+      // - max collPerToken should be point at which the liquidation is invalid
+      //   (i.e. equal to the coll requirement).
+      //   The maximum amount of backing collateral in the position for it to be liquidatable
+      //   is 1.25 x the current price of the token debt, which simplifies to 1.25 `latestPrice`.
+      // - (TODO) If the current ratio of backing collateral to token debt is < `minCollPerToken`
+      //   or > `maxCollPerToken`, then the user shouldn't be liquidating the position.
+      //   Perhaps we should block this functionality in this situation.
```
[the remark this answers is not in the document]
mrice32: Yeah, I was thinking about this too. It seems weird to set the min above where the position currently is or set the max below where it currently is since the txn would be expected to revert.

Agree with you 100%. Maybe (?) it makes sense to:
1. Block if the field contains a value that would cause the transaction to revert.
2. Warn them if they populate the field with a value that will allow the transaction to go through, but we think that they will lose money on the liquidation (either because it will be invalid or there's not enough collateral).

But ^ is just a straw man.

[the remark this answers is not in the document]
nicholaspai: yeah I am leaning more towards 2 but also we should disable the button if the transaction would revert.

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -103,6 +104,40 @@ const PositionActionsDialog = (props: DialogProps) => {
...
+      const _minCollPerTokenToBeProfitable = latestPrice;
+      const _maxCollPerTokenToBeValid =
+        latestPrice * parseFloat(fromWei(collReq));
+
+      setMinCollPerToken(_minCollPerTokenToBeProfitable.toFixed(10));
```
nicholaspai, line 133: I'm thinking rather than round here, I'll just round in the JSX element

features/all-positions/PositionActionsDialog.tsx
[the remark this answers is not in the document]
nicholaspai: I want to leave this for another PR as this likely calls for a signficiant UI overhaul in the `PositionDialog` feature

https://github.com/UMAprotocol/emp-tools/issues/129

PR #128, opened 2020-08-19
feat(All): Add support for non-18-decimal collateral currencies

nicholaspai:
Open questions: 
- What price feeds to use for USDBTC
- How we decide to name y products backed by BTC


containers/EmpSponsors.ts
```diff
@@ -121,7 +122,7 @@ const useEmpSponsors = () => {
           let newLiquidations: LiquidationMap = {};
 
           const collReqFromWei = parseFloat(
-            fromWei(collateralRequirement, collDecs)
+            fromWei(collateralRequirement, tokenDecs)
```
nicholaspai, line 125: This section is still a WIP because the subgraph currently does not account for non 18 decimal precision

features/contract-state/GeneralInfo.tsx
```diff
@@ -64,8 +64,8 @@ const GeneralInfo = () => {
     const expiryDate = new Date(
       expiry.toNumber() * 1000
     ).toLocaleString("en-GB", { timeZone: "UTC" });
-    const prettyLatestPrice = Number(latestPrice).toFixed(4);
-    const pricedGcr = (gcr / latestPrice).toFixed(4);
+    const prettyLatestPrice = Number(latestPrice).toFixed(8);
```
nicholaspai, line 67: This is to account for an identifier like USDBTC where the price is 0.000085... which requires > 4 decimals

features/manage-position/Create.tsx
```diff
@@ -90,7 +90,7 @@ const Create = () => {
     collSymbol !== null &&
     priceIdentifier !== null
   ) {
-    const collReqFromWei = parseFloat(fromWei(collReq, collDec));
+    const collReqFromWei = parseFloat(fromWei(collReq, tokenDec));
```
nicholaspai, line 93: This is a mistake fixed throughout the PR: the collateral requirement unit;s should be the same as the token decimal (or maybe it should be 18) but it unintuitvely should NOT be the collateral's decimal

features/manage-position/Create.tsx
```diff
@@ -144,7 +144,7 @@ const Create = () => {
         setSuccess(null);
         setError(null);
         try {
-          const collateralWei = toWei(collateral);
+          const collateralWei = toWei(collateral, collDec);
```
nicholaspai, line 147: This is the other mistake in this PR: converting collateral token units to Wei requires us to explicitly set the collDec

utils/getOffchainPrice.ts
```diff
@@ -63,6 +71,8 @@ export function getPricefeedParamsFromTokenSymbol(symbol: string | null) {
       return PRICEFEED_PARAMS.compusd;
     case symbol?.includes("ETHBTC"):
       return PRICEFEED_PARAMS.ethbtc;
+    case symbol?.includes("renBTC"):
```
nicholaspai, line 74: Hacky. for now until we determine what a better nomenclature system should be

mrice32:

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -230,7 +234,7 @@ const PositionActionsDialog = (props: DialogProps) => {
         setHash(null);
         setSuccess(null);
         setError(null);
-        const collateralToDepositWei = toWei(collateralToDeposit);
+        const collateralToDepositWei = toWei(collateralToDeposit, collDecs);
```
daywiss, line 237: did not know about this function, that seems useful. so what is this doing exactly? the collateraltodeposit is a number entered by the user as float, and this converts it to wei in its native decimals?

nicholaspai: Yep exactly, where `collDecs = 8` if the collateral currency uses 8 decimal precision

containers/EmpSponsors.ts
```diff
@@ -120,9 +120,7 @@ const useEmpSponsors = () => {
           let newPositions: SponsorMap = {};
           let newLiquidations: LiquidationMap = {};
 
-          const collReqFromWei = parseFloat(
```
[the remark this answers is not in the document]
mrice32: Wait, but the collateral can be non-18 decimal, right? So wouldn't we _want_ to pass in `collDec` to make sure the `fromWei` call uses the correct number of decimals for the conversion? I feel like I'm probably just not seeing something obvious :)

utils/getOffchainPrice.ts
```diff
@@ -52,6 +52,14 @@ export const PRICEFEED_PARAMS: PricefeedParamsMap = {
...
     ],
   },
+  usdbtc: {
+    invertedPrice: true,
+    source: [
+      "https://api.binance.com/api/v3/avgPrice?symbol=BTCUSDT",
+      "https://api.pro.coinbase.com/products/BTC-USD/trades?limit=1",
+      "https://api.kraken.com/0/public/Ticker?pair=BTCUSD",
+    ],
+  },
```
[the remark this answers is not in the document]
mrice32: I think it technically says the quote currency is USD(T). I think this is implying that USDT can be subbed in for USD where a USD trading pair isn't available, but it definitely isn't super clear.

This is a good point, though, and I think we should make the UMIPs more explicit about which pair will be used on which exchanges.

containers/EmpSponsors.ts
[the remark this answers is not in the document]
nicholaspai: I was just following the tests for the EMP which uses 18 decimals to set contract-state variables. I think all contract state values are assumed to be in 18 decimals: https://github.com/UMAprotocol/protocol/blob/master/packages/core/test/financial-templates/Liquidatable.js#L1669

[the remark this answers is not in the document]
nicholaspai: The collateral requirement is only used in Liquidatable's `_settle` method (please correct me if I'm wrong), and it is used to multiply by `tokensOutstanding` to derive the `tokenRedemptionValue`. So I think it actually has to have same precision as `tokensOutstanding`: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol#L526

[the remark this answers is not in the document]
mrice32: Ahhh, sorry for the confusion. You are 100% right. I _was_ missing something obvious haha. This is the collateral _requirement_, which is a percentage (I should have read more carefully). As you said, these sorts of numbers always have a precision of 18 decimals in the contract, since the fixed point library understands 1.2e18 to mean 1.2 (or 120%). Once again, sorry for the confusion. LGTM!

features/manage-position/YourLiquidations.tsx
```diff
@@ -40,13 +39,12 @@ const YourLiquidations = () => {
 
   if (
     tokenSymbol !== null &&
-    collDec !== null &&
     collReq !== null &&
     priceIdentifier !== null &&
     liquidationLiveness !== null &&
     liquidations !== null
   ) {
-    const collReqFromWei = parseFloat(fromWei(collReq, collDec));
```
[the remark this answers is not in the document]
mrice32: Similar to the above, I was totally misunderstanding. This is a percentage, so it should always be 18-decimal denominated. LGTM!

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -298,7 +302,10 @@ const PositionActionsDialog = (props: DialogProps) => {
         setHash(null);
         setSuccess(null);
         setError(null);
-        const collateralToDepositWei = toWei(collateralToDeposit);
+        const collateralToDepositWei = toWei(
```
nicholaspai, line 305: Throughout this PR I round all numbers before calling `toWei` on them because its possible that the string to convert `toWei` has more decimals than the number in `wei` can convert, which throws an error. Example, converting `toWei(0.123456789, 8)` will throw an error because there are 9 decimals but the precision specified is 8.

I hard code 6 as a safe precision value meaning this dApp will work for all tokens with >= 6 precision

utils/convertToWeiSafely.ts
```diff
@@ -0,0 +1,20 @@
+import { utils } from "ethers";
+const { parseUnits: toWei } = utils;
+
+// Minimum precision for a token or collateral currency that this dApp can support
+const MIN_TOKEN_PRECISION = 6;
```
nicholaspai, line 5: Throughout this PR I round all numbers before calling `toWei` on them because its possible that the string to convert `toWei` has more decimals than the number in `wei` can convert, which throws an error. Example, converting `toWei(0.123456789, 8)` will throw an error because there are 9 decimals but the precision specified is 8.

I hard code 6 as a safe precision value meaning this dApp will work for all tokens with >= 6 precision

mrice32: Wait, @nicholaspai why not just use the number of decimals to round? `.toFixed(decimals)`? (I could be missing some context here on why that would be hard/annoying)

nicholaspai: @mrice32 actually I think you can do that

nicholaspai: Fixed

features/all-positions/LiquidationActionDialog.tsx
```diff
@@ -88,7 +88,7 @@ const LiquidationActionDialog = (props: DialogProps) => {
   } = empState;
 
   const prettyBalance = (x: number) => {
-    const x_string = x.toFixed(4);
+    const x_string = x.toFixed(8);
```
chrismaree, line 91: does this not mess up the scaling when `prettyBalance` is used?

utils/convertToWeiSafely.ts
```diff
@@ -0,0 +1,21 @@
...
+// If the argument to pass into `toWei()` has too much precision (specifically more than `precisionToUse`),
+// then `toWei()` might return a string number with decimals, which Solidity cannot handle.
+export function toWeiSafe(
+  numberToConvertToWei: string,
+  desiredPrecision?: number
+) {
+  const precisionToUse = desiredPrecision
+    ? desiredPrecision
+    : DEFAULT_PRECISION;
+  return toWei(
```
chrismaree, line 17: here, where you feed in the precision to use, is this notation only available for `ethers`? can you do this with web3.js as well?

features/manage-position/Withdraw.tsx
```diff
@@ -442,12 +439,12 @@ const Deposit = () => {
       <Box>
         <Box py={2}>
           <Typography>
-            <i>Create a position before depositing more collateral.</i>
+            <i>Create a position before withdrawing collateral.</i>
```
chrismaree, line 442: jeez good catch

features/manage-position/Redeem.tsx
```diff
@@ -103,7 +105,7 @@ const Redeem = () => {
         setSuccess(null);
         setError(null);
         try {
-          const tokensToRedeemWei = toWei(tokens);
+          const tokensToRedeemWei = toWeiSafe(tokens);
```
chrismaree, line 108: I know it defaults but why are you not feeding in decimals here?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -519,8 +527,8 @@ const PositionActionsDialog = (props: DialogProps) => {
                             ? "increase"
                             : "decrease"}{" "}
                           by {Math.abs(underCollateralizedPercent).toFixed(4)}%
-                          from {latestPrice.toFixed(4)} to{" "}
-                          {underCollateralizedPrice.toFixed(4)}. You can still
+                          from {latestPrice.toFixed(8)} to{" "}
```
chrismaree, line 530: why did you make this change?

utils/convertToWeiSafely.ts
nicholaspai: Actually this doesn't work for `web3` because `web3` only allows you to specify `ethers` or `gwei` etc. Look at this PR i made in `protocol` where we had to import part of the ethers package to get this functionality for our `web3` repo: https://github.com/UMAprotocol/protocol/pull/1899

features/manage-position/Redeem.tsx
nicholaspai: Ah great catch thanks

features/all-positions/PositionActionsDialog.tsx
nicholaspai: Because for example the price fo USDBTC is 0.00008X which would just be 0.0004 if only 4 deciamls. Admittedly hacky but we should fix better in future PR

features/all-positions/LiquidationActionDialog.tsx
nicholaspai: It does but for USDBTC we need more than 4 decimals....

PR #132, opened 2020-08-25
feat(balancer-container): Auto detect relevant Balancer pool to query data from

nicholaspai:
Checks if user has selected a yToken against a list of yield tokens, default is the yUSD-SEP20 token.

Queries balancer pool data from yield token.

Also adds information about farming UMA during roll between SEPT and OCT

Preview prior to roll on august 28 23:00 UTC:
Screen Shot 2020-08-25 at 17 19 13

Preview after roll:
Screen Shot 2020-08-25 at 17 19 40

### DIFF
--- a/containers/Balancer.ts
+++ b/containers/Balancer.ts
@@ -44,7 +44,7 @@ const useBalancer = () => {
 
   // Detect if selected token is a yield token, otherwise default to a yield token. This component
   // does not make sense for a non yield token.
-  const defaultTokenAddress = "0x81ab848898b5ffd3354dbbefb333d5d183eedcb5";
+  const defaultTokenAddress = YIELD_TOKENS[0].toLowerCase();
   const usdcAddress = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48";
   const defaultSwapTokenAddress = usdcAddress;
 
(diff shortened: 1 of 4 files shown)


features/yield/Yield.tsx
```diff
@@ -57,6 +57,14 @@ const Yield = () => {
             >
               here
             </a>
+            , and to learn about rolling between expiries see the article{" "}
```
chrismaree, line 60: +1

chrismaree (approved):
this looks like a pretty good start! I think that you should hide the yield tab if you are on a non-yield token though. for example I can select ethbtc & still see a price, which is confusing. Once you've changed this I think we are good to go.

nicholaspai:
> this looks like a pretty good start! I think that you should hide the yield tab if you are on a non-yield token though. for example I can select ethbtc & still see a price, which is confusing. Once you've changed this I think we are good to go.

Agreed with this. I implemented pls play around!

containers/Balancer.ts
```diff
@@ -50,30 +73,38 @@ const useBalancer = () => {
...
   } = useQuery(TOKENS, {
+    skip: !selectedTokenAddress,
     context: { clientName: "BALANCER" },
-    variables: { tokenId: empAddress.toLowerCase() },
+    variables: { tokenId: selectedTokenAddress },
     pollInterval: 5000,
   });
   const { loading: poolLoading, error: poolError, data: poolData } = useQuery(
-    POOL(JSON.stringify(poolTokensList.map((token) => token.toLowerCase()))),
+    POOL(JSON.stringify(poolTokenList)),
```
nicholaspai, line 82: I set these queried addresses explicitly to lower case below when I set state

mrice32:

chrismaree:
This is great! you new changes are perfect. thanks for adding this @nicholaspai . I'm going to merge this on your behalf because people are asking about this feature in discord.

PR #136, opened 2020-08-26
feat(click-to-liquidate): Block invalid min/maxCollPerToken

### LINKED ISSUE #129: Liquidation dialog: make it more clear what reasonable values for  `minCollperToken` and `maxCollperToken would be `
`minCollPerToken` should be the minimum ratio of collateral/token in a position such that liquidating it is profitable

`maxCollPerToken` should be the maximum ratio of collaterall/token in a position such that liquidating it won't get disputed

We should potentially disable the click to liquidate button if the position's collateral/token is > `maxCollPerToken` (as  or < `minCollPerToken` as defined above

nicholaspai:
Fixes #129 

Some previews:

New rows showing current, min, and max collateral ratios:
Screen Shot 2020-08-26 at 09 33 41

Warning user if they set maxCollPerToken too high:
Screen Shot 2020-08-26 at 09 33 51

Warning a user if they set minCollPerToken too low:
Screen Shot 2020-08-26 at 09 34 06

Error shown to user if the position's cratio is outside their specified min/max bounds:
Screen Shot 2020-08-26 at 09 43 35

Rendering shown if only one bound is exceeded:
Screen Shot 2020-08-26 at 09 43 43
Screen Shot 2020-08-26 at 09 43 53

### DIFF
--- a/features/all-positions/PositionActionsDialog.tsx
+++ b/features/all-positions/PositionActionsDialog.tsx
@@ -520,18 +520,49 @@ const PositionActionsDialog = (props: DialogProps) => {
                           price.
                           <br></br>
                           <br></br>
-                          <Status>
-                            <Label>Current collateral/token ratio: </Label>
-                            {collRatio.toFixed(4)}
-                          </Status>
-                          <Status>
-                            <Label>Minimum profitable ratio: </Label>
-                            {minCollPerTokenToBeProfitable}
-                          </Status>
-                          <Status>
-                            <Label>Maximum liquidatable ratio: </Label>
-                            {maxCollPerTokenToBeValid}
-                          </Status>
+                          <Tooltip
+                            placement="top"
+                            title={`This can be calculated as (amount of collateral) divided by 
+                              (amount of tokens)`}
+                          >
+                            <Status>
+                              <Label>Current collateral/token ratio: </Label>
+                              {collRatio.toFixed(4)}
+                            </Status>
+                          </Tooltip>
+                          <Tooltip
+                            placement="top"
+                            title={`Liquidating a position whose collateral ratio is
+                              below ${minCollPerTokenToBeProfitable} would not be
+                              profitable using the estimated 
+                              ${utils.parseBytes32String(
+                                priceId
+                              )} price. This is
+                              because you must burn synthetic tokens in order to
+                              liquidate underlying collateral, and we estimate
+                              that each ${tokenSymbol} is worth
+                              ${latestPrice.toFixed(4)} of the collateral
+                              ${collSymbol}.`}
+                          >
+                            <Status>
+                              <Label>Minimum profitable ratio: </Label>
+                              {minCollPerTokenToBeProfitable}
+                            </Status>
+                          </Tooltip>
+                          <Tooltip
+                            placement="top"
+                            title={`Liquidating a position whose collateral ratio is
+                              greater than ${maxCollPerTokenToBeValid} would get
+                              disputed using the estimated
+                              ${utils.parseBytes32String(priceId)} price of 
+                              ${latestPrice.toFixed(4)} and the collateral
+                              requirement of ${parseFloat(fromWei(collReq))}.`}
+                          >
+                            <Status>
+                              <Label>Maximum liquidatable ratio: </Label>
+                              {maxCollPerTokenToBeValid}
+                            </Status>
+                          </Tooltip>
                         </Typography>
                         {minCollPerTokenNum <
                           minCollPerTokeToBeProfitablenNum && (
--- a/features/all-positions/PositionActionsDialog.tsx
+++ b/features/all-positions/PositionActionsDialog.tsx
@@ -529,7 +529,7 @@ const PositionActionsDialog = (props: DialogProps) => {
                           <Tooltip
                             placement="top"
                             title={`This can be calculated as (amount of collateral) divided by 
-                              (amount of tokens)`}
+                              (amount of tokens)`}.
                           >
                             <Status>
                               <Label>Current collateral/token ratio: </Label>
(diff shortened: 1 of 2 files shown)


features/all-positions/PositionActionsDialog.tsx
```diff
@@ -318,6 +332,7 @@ const PositionActionsDialog = (props: DialogProps) => {
       }
     };
 
+    console.log(Number(minCollPerToken), minCollPerTokenToBeProfitable);
```
kendricktan, line 335: Might wanna remove this console.log

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -489,19 +504,69 @@ const PositionActionsDialog = (props: DialogProps) => {
...
-                          <a
-                            href={DOCS_MAP.FINAL_FEE}
-                            target="_blank"
-                            rel="noopener noreferrer"
-                          >
-                            UMA docs
-                          </a>
-                          .
-                        </Important>
+                        {Number(minCollPerToken) <
```
kendricktan, line 520: There's a couple of `Number(minCollPerToken)` thoughout the place. Seems like its converting it from a string. Should we take into account that its a NaN?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -466,7 +481,7 @@ const PositionActionsDialog = (props: DialogProps) => {
                               <br></br>
                               <strong>Note:</strong> You will need to sign two
                               transactions one to approve {collSymbol} and a
-                              second to preform the deposit.
+                              second to perform the deposit.
```
chrismaree, line 484: good catch 😅

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -523,6 +595,13 @@ const PositionActionsDialog = (props: DialogProps) => {
...
+                                error={collRatio > Number(maxCollPerToken)}
+                                helperText={
+                                  collRatio > Number(maxCollPerToken) &&
+                                  `Current collateral ratio of ${collRatio.toFixed(
+                                    4
+                                  )} is above the specified maximum ratio, transaction will revert.`
+                                }
                                 onChange={(
                                   e: React.ChangeEvent<HTMLInputElement>
                                 ) => setMaxCollPerToken(e.target.value)}
```
chrismaree, line 607: do you want to perhaps auto-fill the max tokens to liquidate as the full position size?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -523,6 +595,13 @@ const PositionActionsDialog = (props: DialogProps) => {
                                 inputProps={{ min: "0" }}
```
chrismaree, line 595: this input prop we might want to change. It makes this UI element pretty hard to use. I tried to increment the `mincollperToken` by pressing the up arrow but it increments in `1` each time. 



How do you feel about calculating this prop such that it increments by some logical amount relative to the `Max collateral/token` size?

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -489,19 +504,69 @@ const PositionActionsDialog = (props: DialogProps) => {
                           price.
                           <br></br>
                           <br></br>
+                          <Status>
+                            <Label>Current collateral/token ratio: </Label>
```
chrismaree, line 508: This table is super useful. However, I think it's pretty complex unless you understand the details of our liquidation mechanism. Could we add some hover information explaining in concise terms that explains what each of these values are? 

The same could be said for the 4 input properties below. For those we could even take the contract natspec directly from the solidity files.

features/all-positions/PositionActionsDialog.tsx
nicholaspai: Great catch, I initialized new variables before the render like so:

```
const minCollPerTokenNum = Number(minCollPerToken) || 0;
```

features/all-positions/PositionActionsDialog.tsx
nicholaspai: What if user doesn't have enough balance?

nicholaspai: I auto filled to max
Screen Shot 2020-08-27 at 08 17 20

features/all-positions/PositionActionsDialog.tsx
nicholaspai: great idea! I changed to `step: (collRatio / 100)`

features/all-positions/PositionActionsDialog.tsx
nicholaspai: I added tool tip above the table:
Screen Shot 2020-08-27 at 08 27 22

features/all-positions/PositionActionsDialog.tsx
mrice32: We could do the min of the position size and their balance?

nicholaspai: Yeah @mrice32 that's easiest implementation I think

features/all-positions/PositionActionsDialog.tsx
```diff
@@ -489,28 +526,115 @@ const PositionActionsDialog = (props: DialogProps) => {
...
-                          Exercise caution! Incorrectly liquidating a position
-                          can lose you money! See the{" "}
-                          <a
-                            href={DOCS_MAP.FINAL_FEE}
-                            target="_blank"
-                            rel="noopener noreferrer"
+                          <Tooltip
+                            placement="top"
+                            title={`This can be calculated as (amount of collateral) divided by 
+                              (amount of tokens)`}
```
chrismaree, line 532:
```suggestion
                              (amount of tokens)`}.
```

chrismaree (approved):
This is awesome! thanks for refining this component. it's way more useable now.

PR #137, opened 2020-08-26
Add copy-to-clipboard button and etherscan link for EMP address, and add github link to footer

### LINKED ISSUE #76: Add helpful links to footer
Some candidate helpful links (Wrapped in IconButtons):
- Smart contract source code
- dApp source code
- EMP smart contract audit by OpenZeppelin-
- Community audit
- UMA docs

nicholaspai:
Fixes #76 
Screen Shot 2020-08-26 at 11 39 54


chrismaree (approved):
looks good! 

are there more links and icons we'd want to add to the footer? discord, twitter, main website? many other things we could add if we wanted to

mrice32:

nicholaspai:
> looks good!
> 
> are there more links and icons we'd want to add to the footer? discord, twitter, main website? many other things we could add if we wanted to

I added Discord and Twitter link. The icon sizes don't look quite right to me. Do you know enough CSS to make them all the same size? Its a bit tricky because github is a MUI icon while discord and twitter are SVG's, but we can also switch Github to an svg

mrice32:
> > looks good!
> > are there more links and icons we'd want to add to the footer? discord, twitter, main website? many other things we could add if we wanted to
> 
> I added Discord and Twitter link. The icon sizes don't look quite right to me. Do you know enough CSS to make them all the same size? Its a bit tricky because github is a MUI icon while discord and twitter are SVG's, but we can also switch Github to an svg

I also noticed that the Github button lights up red when you hover but the other two don't change colors. I think the sizing, etc is probably fine for now. Not something worth spending a lot of time on. Something for a seasoned UI engineer to figure out :)

pages/index.tsx
```diff
@@ -22,6 +24,11 @@ import AllPositions from "../features/all-positions/AllPositions";
 import Weth from "../features/weth/Weth";
 import Yield from "../features/yield/Yield";
 
+import GitHubIcon from "@material-ui/icons/GitHub";
+import DiscordIcon from "../public/discord.svg";
+import TwitterIcon from "../public/twitter.svg";
```
chrismaree, line 29: I'm not really a fan of having to import svg icons. ideally, we should only use stuff from `@material-ui/icons/GitHub`. 

Twitter has an icon:
```
import TwitterIcon from '@material-ui/icons/Twitter';
```

Discord doesn't have one though so I guess we'll have to use that from an SVG.

pages/index.tsx
```diff
@@ -22,6 +24,11 @@ import AllPositions from "../features/all-positions/AllPositions";
...
+import VercelIcon from "../public/vercel.svg";
```
chrismaree, line 30: Also, I'm pretty sure we are not using the vercal Icon?

nicholaspai: Yeah we're not my bad

pages/index.tsx
nicholaspai: We shouoldn't mix and match SVG vs MUI Icon as they look different, so I propose we either move all to SVG or all to MUI Icon (and remove discord)

nicholaspai: I'm going to go ahead and remove discord as the twitter bio has a discord link

nicholaspai: I removed it it looks much better now PTAL

mrice32: That sounds good. At some point down the road, we might want to have some sort of link somewhere else on the page:
```
Having issues? Use our [DISCORD_LINK] to get help!"
```

PR #139, opened 2020-08-29
fix(general): two nit fixes

nicholaspai:
- reduce "max" tokens in Create so that GCR constraint is always met
- Hide countdown to roll when time is up

### DIFF
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -96,8 +96,10 @@ const Create = () => {
     // is intended to encourage users to maximize their capital efficiency.
     const maxTokensToCreate = _gcr > 0 ? collateral / _gcr : 0;
     // Unlike the min collateral, we're ok if we round down the tokens slightly as round down
-    // can only increase the position's CR and maintain it above the GCR constraint
-    setTokens(maxTokensToCreate.toFixed(4));
+    // can only increase the position's CR and maintain it above the GCR constraint. However, we'll
+    // subtract a tiny amount of tokens because the prevision division might unintentionally floor
+    // the quotient.
+    setTokens((maxTokensToCreate - 1).toFixed(4));
   };
 
   if (
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -96,10 +96,8 @@ const Create = () => {
     // is intended to encourage users to maximize their capital efficiency.
     const maxTokensToCreate = _gcr > 0 ? collateral / _gcr : 0;
     // Unlike the min collateral, we're ok if we round down the tokens slightly as round down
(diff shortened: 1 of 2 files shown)


features/manage-position/Create.tsx
```diff
@@ -96,8 +96,10 @@ const Create = () => {
     // is intended to encourage users to maximize their capital efficiency.
     const maxTokensToCreate = _gcr > 0 ? collateral / _gcr : 0;
     // Unlike the min collateral, we're ok if we round down the tokens slightly as round down
-    // can only increase the position's CR and maintain it above the GCR constraint
-    setTokens(maxTokensToCreate.toFixed(4));
+    // can only increase the position's CR and maintain it above the GCR constraint. However, we'll
+    // subtract a tiny amount of tokens because the prevision division might unintentionally floor
+    // the quotient.
+    setTokens((maxTokensToCreate - 1).toFixed(4));
```
mrice32, line 102: Why do you have to subtract 1? Can’t you just subtract like 0.00005 or something?

nicholaspai: the current check for enabling the Create button is:

`transactionCR > GCR` where `transactionCR = collateral / tokens` and `collateral` and `tokens` are the string values (not numbers) set by these two `setStates` from Min and Max functions. So this is by far the easiest way to do it without changing other logic (which we probably should but I wanted to keep out of scope).

This admittedly won't scale for certain types of token where 1 synthetic represents a lot of real value.

mrice32: In the interest of speed, I think this is good for now!

I don't quite understand why we would subtract from the quotient if we're worried the division may have floored it. I would imagine that we'd want to _add_ in that case. I also don't see why floating point division would ever ceil or floor unless it lost precision. In the case of lost precision, I'm actually not too sure whether we have _any_ guarantees about the error without knowing how large the numbers are.

nicholaspai:
My comments actually incorrect. The .toFixed(4) floors it and transactionCR computes using those floored values. So a better solution might be to subtract a smaller amount like 0.0005

features/manage-position/Create.tsx
```diff
@@ -96,7 +96,7 @@ const Create = () => {
     // is intended to encourage users to maximize their capital efficiency.
     const maxTokensToCreate = _gcr > 0 ? collateral / _gcr : 0;
     // Unlike the min collateral, we're ok if we round down the tokens slightly as round down
-    // can only increase the position's CR and maintain it above the GCR constraint
+    // can only increase the position's CR and maintain it above the GCR constraint.
```
nicholaspai, line 99: @mrice32 I removed the rounding, it seems to work

PR #140, opened 2020-08-29
feat(emp-analitics): Initial plots

chrismaree:
This PR adds some initial analytics to emp-tools that lets you see the `EMP Token Sponsor Liquidation Prices` and the `Current Balancer Liquidity Providers` plots. This PR uses apex charts and for now does not require any additional data to be pulled into the front end (everything we need is there already). 

image

image

There are many other charts We can add into the future. The next plot I want to generate is historic yUSD pool sizes with both pool super imposed. With this plot you’ll be able to see the progress of a given roll. Potentially also would like to plot the yUSD token price over time in a separate plot. Also we could add automatic breakdowns for all the LM rewards for each week in a plot.


features/yield/Yield.tsx
```diff
@@ -34,8 +34,8 @@ const Yield = () => {
       <Box py={2}>
         <Typography>
           <i>
-            Please first connect and set your network to Mainnet, and then
-            select a yield token (i.e. yUSD).
+            Please first connect to Mainnet, and then select a yield token (i.e.
```
chrismaree, line 37: changed this so it would not wrap to two lines on normal screen size.

nicholaspai:
@chrismaree this is going to be very useful! One suggestion i have is for the liquidity providers graph to show a [cumulative distribution graph](https://en.wikipedia.org/wiki/Cumulative_distribution_function) of liquidity providers, instead of amount per account. This would be useful to know how much liquidity is supplied by the top 10% largest LP's. 

Maybe it doesn't actually have to replace the "addresses" as the x-axis, but you can place vertical lines that show the 25th, 50th, 75th percentile of liquidity provided is.

Another question I have is, what does "logarithmic" mean for the Liquidity price graph? The graph looks a lot nicer but I don't understand intuitively what "logarithmic" liquidity price means

mrice32:
@chrismaree do you want reviewers on this now?

pages/index.tsx
```diff
@@ -35,6 +37,7 @@ import TwitterIcon from "@material-ui/icons/Twitter";
 const StyledTabs = styled(Tabs)`
   & .MuiTabs-flexContainer {
     border-bottom: 1px solid #999;
+    width: 200%;
```
chrismaree, line 40: this is added to ensure that the underline in the component extends far enough to accommodate many tabs in the case of including the analytics tab & weth.

pages/index.tsx
```diff
@@ -48,44 +51,50 @@ const Blurb = styled.div`
 `;
 
 export default function Index() {
-  const [tabIndex, setTabIndex] = useState(0);
```
chrismaree, line 51: I had to pretty much re-write the logic in the menu to accommodate the dynamic options depending on the EMP selected. Also, there was some inconsistent behavior when switching between EMPs that offer different tabs. Now everything works from the `selectedMenuItem` string.

features/analytics/CurrentSponsorsCollateralRatio.tsx
```diff
@@ -0,0 +1,154 @@
...
+        },
+        yaxis: {
+          title: {
+            text: `Position Collateral (${collSymbol})`,
+            offsetY: 15,
+          },
+        },
+        plotOptions: {
+          bar: {
+            columnWidth: Number(prettyLatestPrice) / 3, // scale the column width as a function of price.
```
chrismaree, line 101: this keeps the bar width a function of the x-axis maximum scale.

chrismaree:
> @chrismaree do you want reviewers on this now?

it is now :)

chrismaree:
> @chrismaree this is going to be very useful! One suggestion i have is for the liquidity providers graph to show a [cumulative distribution graph](https://en.wikipedia.org/wiki/Cumulative_distribution_function) of liquidity providers, instead of amount per account. This would be useful to know how much liquidity is supplied by the top 10% largest LP's.
> 
> Maybe it doesn't actually have to replace the "addresses" as the x-axis, but you can place vertical lines that show the 25th, 50th, 75th percentile of liquidity provided is.
> 
> Another question I have is, what does "logarithmic" mean for the Liquidity price graph? The graph looks a lot nicer but I don't understand intuitively what "logarithmic" liquidity price means

shall we leave this for a future version?

daywiss (approved):

pages/index.tsx
```diff
@@ -48,44 +51,50 @@ const Blurb = styled.div`
...
+      options = options.concat(["yUSD Yield", "Analytics"]);
+    }
+
+    // Update selected page if the user toggles between EMPs while selected on
+    // invalid pages (i.e on Wrap/Unwrap then moves to uUSDrBTC).
+    if (options.indexOf(selectedMenuItem) == -1) {
+      setSelectedMenuItem("General Info");
+    }
+  };
+  buildOptionsList();
```
mrice32, line 85: Why do you call this in the mainline render _and_ in the use effect? I haven't deeply absorbed the flow, but this just struck me as a little off.

pages/index.tsx
```diff
@@ -48,44 +51,50 @@ const Blurb = styled.div`
...
+  let options = ["General Info", "Manage Position", "All Positions"];
+  const buildOptionsList = () => {
+    // If it is weth collateral contract then add the weth option.
+    if (weth && collAddress?.toLowerCase() == weth.address.toLowerCase()) {
+      options.push("Wrap/Unwrap WETH");
+    }
+
+    // If it is a yield token then add the yUSD yield and analytics tabs.
+    if (isYieldToken) {
+      options = options.concat(["yUSD Yield", "Analytics"]);
```
mrice32, line 76: What's the definition of a yield token? Just a USD tracking one? Does the analytics tab still make sense on a non-usd tracking token that has liquidity mining? What about USD-tracking tokens that have no liquidity mining?

mrice32 (commented):
Looks great! A few minor questions and comments.

features/analytics/CurrentLiquidityProviders.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import dynamic from "next/dynamic";
+const Chart = dynamic(() => import("react-apexcharts"), { ssr: false });
+
+const prettyAddress = (x: string) => {
+  return x.substr(0, 4) + "..." + x.substr(x.length - 4, x.length);
+};
+
+const CurrentLiquidityProviders = () => {
+  const { pool, shares } = Balancer.useContainer();
```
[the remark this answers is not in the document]
nicholaspai: Can disregard now

pages/index.tsx
nicholaspai: Its any token found in this [dictionary](https://github.com/UMAprotocol/emp-tools/blob/master/constants/yieldTokens.ts#L11) which is used to determine when to render the Yield calculator tab for example

features/analytics/CurrentLiquidityProviders.tsx
```diff
@@ -0,0 +1,106 @@
...
+    });
+
+    const shareArraySorted = shareArray.sort((a, b) =>
+      a.balance < b.balance ? 1 : -1
+    );
+    const totalBPT = shareArray.reduce((a, b) => {
+      return a + b.balance;
+    }, 0);
+
+    const usdPerBPT = pool.liquidity / totalBPT;
```
nicholaspai, line 38: Should you check for divide by 0?

pages/index.tsx
```diff
@@ -48,44 +51,50 @@ const Blurb = styled.div`
...
-  if (weth && collAddress?.toLowerCase() == weth.address.toLowerCase()) {
-    options.push("Wrap/Unwrap WETH");
-  }
-
-  // Update selected page if the user toggles between EMPs while selected on
-  // invalid pages (i.e on Wrap/Unwrap then moves to uUSDrBTC)
-  if (tabIndex > options.length - 1) {
-    setTabIndex(0);
-  }
+  useEffect(() => {
```
nicholaspai, line 87: Adding a `useEffect` to the main `pages/index.tsx` feels risky performance-wise for me. Have you stress tested the `buildOptionsList` method? Just want to double check here

features/analytics/CurrentSponsorsCollateralRatio.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+import EmpState from "../../containers/EmpState";
+import Collateral from "../../containers/Collateral";
+import EmpSponsors from "../../containers/EmpSponsors";
+import PriceFeed from "../../containers/PriceFeed";
+import Token from "../../containers/Token";
+import { isPricefeedInvertedFromTokenSymbol } from "../../utils/getOffchainPrice";
+
+import dynamic from "next/dynamic";
+const Chart = dynamic(() => import("react-apexcharts"), { ssr: false });
```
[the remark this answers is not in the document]
chrismaree: Next provides structure for building your app and comes prebundled with routing and code spliting. The main feature of Next is Server Side Rendering(SSR) allowing your site/app's HTML to be rendered on the server and displayed in your browser. Sometimes there are packages and/or libraries we would want to include in our application that cannot be rendered on the server. With these pieces of code, we need let the browser do the rendering. The `react-apexcharts` need to be client-side only; this import syntax disables the SSR on this package.

I've added a comment to the code to explain this for future reference.

features/analytics/CurrentSponsorsCollateralRatio.tsx
```diff
@@ -0,0 +1,154 @@
...
+        const fieldValueA = activeSponsors[sponsorA].liquidationPrice;
+        const fieldValueB = activeSponsors[sponsorB].liquidationPrice;
+        return Number(fieldValueA) > Number(fieldValueB) ? 1 : -1;
+      });
+
+    const sponsorArray = reformattedSponsorKeys.map((address) => {
+      return {
+        address: prettyAddress(address),
+        bucket:
+          Math.ceil(Number(activeSponsors[address].liquidationPrice) / 5) * 5,
```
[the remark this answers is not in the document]
chrismaree: done

pages/index.tsx
chrismaree: Ye, good point. I've refactored this. it's cleaner now.

pages/index.tsx
chrismaree: haven't stress tested it too aggressively tbh but It feels like it's ok. the way it's set up now it will only get called when you switch between EMPs. Previously we were doing way more heavy rendering in this component to build this menu component.

nicholaspai (approved):
I played around on the preview deployment site, nothing breaks and charts work as expected. LGTM

PR #141, opened 2020-08-31
feat(request withdrawal) Added progress bar

chrismaree:
This small PR adds a little feature to the request withdrawal page to show the progress of a withdrawal request.

image

### DIFF
--- a/features/manage-position/Withdraw.tsx
+++ b/features/manage-position/Withdraw.tsx
@@ -125,9 +125,11 @@ const Deposit = () => {
     const hasPendingWithdraw = pendingWithdraw === "Yes";
     const pendingWithdrawTimeRemaining = withdrawPassTime - Number(currentTime);
     const progressBarPercent =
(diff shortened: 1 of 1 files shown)


features/manage-position/Withdraw.tsx
```diff
@@ -123,6 +124,10 @@ const Deposit = () => {
     );
     const hasPendingWithdraw = pendingWithdraw === "Yes";
     const pendingWithdrawTimeRemaining = withdrawPassTime - Number(currentTime);
+    const progressBarPercent =
```
nicholaspai, line 127: What if `withdrawlLiveness` is 0 will this crash?

chrismaree: it doesent crash it but I agree it should not calculate this if it is zero. Have updated PTAL.

chrismaree:
@nicholaspai ptal at this. small pr that we should be able to get in.

nicholaspai (approved):
LGTM

PR #142, opened 2020-09-01
fix(Create): floor max tokens to create instead of rounding to arbitrary precision

nicholaspai:
Avoids edge case where `max` sets tokens to a CR === GCR, `Math.floor(x - some_small_amount)` is cleaner.

### DIFF
--- a/features/manage-position/Create.tsx
+++ b/features/manage-position/Create.tsx
@@ -97,7 +97,7 @@ const Create = () => {
     const maxTokensToCreate = _gcr > 0 ? collateral / _gcr : 0;
     // Unlike the min collateral, we're ok if we round down the tokens slightly as round down
     // can only increase the position's CR and maintain it above the GCR constraint.
-    setTokens(maxTokensToCreate.toFixed(4));
+    setTokens(Math.floor(maxTokensToCreate).toString());
   };
 
   if (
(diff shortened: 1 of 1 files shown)


chrismaree (commented):
how does this scale with non-18 decimals? or is this number input here a non-wei delimited so it's 0.0004 synthetic, irrespective of the calling?

nicholaspai:
> how does this scale with non-18 decimals? or is this number input here a non-wei delimited so it's 0.0004 synthetic, irrespective of the calling?

Yeah it's non wei delimited.

features/manage-position/Create.tsx
```diff
@@ -97,7 +97,7 @@ const Create = () => {
     const maxTokensToCreate = _gcr > 0 ? collateral / _gcr : 0;
     // Unlike the min collateral, we're ok if we round down the tokens slightly as round down
     // can only increase the position's CR and maintain it above the GCR constraint.
-    setTokens(maxTokensToCreate.toFixed(4));
+    setTokens(Math.floor(maxTokensToCreate - 0.00005).toString());
```
mrice32, line 100: Why can't we just do something like this instead:
```
setTokens((maxTokensToCreate - 0.0001).toFixed(4));
```

nicholaspai: We can do this

daywiss (approved):

PR #151, opened 2020-09-07
fix(yield-calculator) fix pool inconsistancies

chrismaree:
This PR fixes an issue in the yield calculator for renBTC. Before this PR, the balancer graphQL query expects the `tokensList` in the pools had the synth token USDC token address in an array. This works for `yUSD-SEP20` and `yUSD-OCT20`. However for `uUSDrBTC-OCT` this order has flipped around. To accommodate any direction, this PR modifies the `YIELD_TOKENS` object to store a string-array key-value object to store relevant balancer lookups with the correct token ordering.

### DIFF
--- a/containers/Balancer.ts
+++ b/containers/Balancer.ts
@@ -73,7 +73,7 @@ const useBalancer = () => {
   const [selectedSwapTokenAddress, setSelectedSwapTokenAddress] = useState<
     string | null
   >(null);
-  const [poolTokenList, setPoolTokenList] = useState<string[] | null>();
+  const [poolTokenList, setPoolTokenList] = useState<string[] | null>(null);
   const [isYieldToken, setIsYieldToken] = useState<boolean>(false);
 
   const [usdPrice, setUsdPrice] = useState<number | null>(null);
--- a/containers/Balancer.ts
+++ b/containers/Balancer.ts
@@ -33,22 +33,29 @@ interface PoolTokenQuery {
   balance: string;
 }
 
-interface YIELD_TOKEN {
-  [key: string]: string[];
+interface yieldPair {
+  [key: string]: string;
 }
-const YIELD_TOKENS: YIELD_TOKEN = {
-  "0x81ab848898b5ffD3354dbbEfb333D5D183eEDcB5": [
-    "0x81ab848898b5ffD3354dbbEfb333D5D183eEDcB5".toLowerCase(),
-    "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48".toLowerCase(),
-  ], // yUSDETH-SEP20
-  "0xB2FdD60AD80ca7bA89B9BAb3b5336c2601C020b4": [
-    "0xb2fdd60ad80ca7ba89b9bab3b5336c2601c020b4".toLowerCase(),
-    "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48".toLowerCase(),
-  ], // yUSDETH-Oct20
-  "0x208D174775dc39fe18B1b374972F77ddEc6c0F73": [
-    "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48".toLowerCase(),
-    "0x208d174775dc39fe18b1b374972f77ddec6c0f73".toLowerCase(),
-  ], // uUSDrBTC-OCT
+
+interface yieldToken {
+  [key: string]: yieldPair;
+}
+
+// The keys in this object are the synthetic token. the `token0` and `token1` are
+// the balancer pool key value pairs for the first and second token in the pool.
+const YIELD_TOKENS: yieldToken = {
+  "0x81ab848898b5ffD3354dbbEfb333D5D183eEDcB5": {
+    token0: "0x81ab848898b5ffD3354dbbEfb333D5D183eEDcB5",
+    token1: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
+  }, // yUSDETH-SEP20
+  "0xB2FdD60AD80ca7bA89B9BAb3b5336c2601C020b4": {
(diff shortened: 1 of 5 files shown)


containers/Balancer.ts
```diff
@@ -33,10 +33,23 @@ interface PoolTokenQuery {
...
 
-const YIELD_TOKENS = [
-  "0x81ab848898b5ffD3354dbbEfb333D5D183eEDcB5", // Sep20
-  "0xB2FdD60AD80ca7bA89B9BAb3b5336c2601C020b4", // Oct20
-];
+interface YIELD_TOKEN {
+  [key: string]: string[];
+}
+const YIELD_TOKENS: YIELD_TOKEN = {
+  "0x81ab848898b5ffD3354dbbEfb333D5D183eEDcB5": [
```
mrice32, line 40: nit: why to lower case instead of to checksum address?

containers/Balancer.ts
```diff
@@ -33,10 +33,23 @@ interface PoolTokenQuery {
...
+    "0x81ab848898b5ffD3354dbbEfb333D5D183eEDcB5".toLowerCase(),
+    "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48".toLowerCase(),
+  ], // yUSDETH-SEP20
+  "0xB2FdD60AD80ca7bA89B9BAb3b5336c2601C020b4": [
+    "0xb2fdd60ad80ca7ba89b9bab3b5336c2601c020b4".toLowerCase(),
```
mrice32, line 45: What are the two elements in these arrays? Would it be clearer to make them key value pairs?

containers/Balancer.ts
chrismaree: the input to the graphQL query only accepts lower case addresses

containers/Balancer.ts
nicholaspai: Are they [0] = synth, [1] = collateral?

chrismaree: In this, the key is the synthetic token and the value is the associated balancer pool pair in the correct order for querying. It cant be a key-value pair as sometimes the ordering in the query within balancer swaps direction. For some of our tokens, this is `[syntheticAddress,USDCAddress]` buy for others, like ren this is `[USDCAddress,syntheticAddress]`.

does that make sense?

features/yield/BalancerData.tsx
```diff
@@ -60,7 +62,7 @@ const BalancerData = () => {
     return (
       <span>
         <Typography variant="h5" style={{ marginBottom: "10px" }}>
-          yUSD Pool Metrics{" "}
+          {tokenSymbol} Pool Metrics{" "}
```
nicholaspai, line 65: Does this render something strange if the selected token is a non yield token?

containers/Balancer.ts
chrismaree: > Are they [0] = synth, [1] = collateral?

not always; this is the reason I had to add this mapping object like this. For the ren synthetic a query that works looks like this:
```
{
    pools(where: {tokensList: ["0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48","0x208d174775dc39fe18b1b374972f77ddec6c0f73"]}) {
      swapFee
      liquidity
      totalSwapVolume
      tokens {
        address
        balance
      }
      joinsCount
      exitsCount
      swapsCount
      totalShares
      shares(first:1000, where: { balance_gt: 0}) {
        userAddress {
          id
        }
        balance
      }
    }
  }
```
This has `[USDCAddress,syntheticAddress]` in the `where` clause. For the other yield tokens this would be swapped.

features/yield/BalancerData.tsx
chrismaree: Balancer data should only show up for a yield token. This page wont load if the token is not a yield token.

containers/Balancer.ts
nicholaspai: This makes sense. Would it be accurate (and more readable) to make the keys `token0` and `token1`?

features/yield/YieldCalculator.tsx
```diff
@@ -58,8 +60,7 @@ const YieldCalculator = () => {
 
     // `yieldPerUnit` = (FACE/yUSD_PX)^(1/(365/DAYS_TO_EXP)) - 1,
     // where FACE = $1. More details: https://www.bankrate.com/glossary/a/apy-annual-percentage-yield/
-    const yieldPerUnit =
-      Math.pow(1 / _tokenPrice, 1 / (_daysToExpiry / DAYS_PER_YEAR)) - 1;
+    const yieldPerUnit = (1 - _tokenPrice) * (DAYS_PER_YEAR / _daysToExpiry);
```
nicholaspai, line 63: @chrismaree @hlambur can you take a quick look at this? Chris thinks that we are calculating APY incorrectly

containers/Balancer.ts
chrismaree: Ye, sure. done. PTAL

containers/Balancer.ts
chrismaree: I cleaned this up a bit by pulling toLowerCase to the set function.

containers/Balancer.ts
```diff
@@ -53,7 +73,7 @@ const useBalancer = () => {
   const [selectedSwapTokenAddress, setSelectedSwapTokenAddress] = useState<
     string | null
   >(null);
-  const [poolTokenList, setPoolTokenList] = useState<string[] | null>(null);
+  const [poolTokenList, setPoolTokenList] = useState<string[] | null>();
```
nicholaspai, line 76: why did you change this?

nicholaspai (approved):
LGTM, I only had 1 question which might be really obvious but I don't udnerstand why you made!

containers/Balancer.ts
chrismaree: have reverted this

PR #160, opened 2020-09-09
fix(farming-calc) APR should not depend on pool share, refactor reward token object

nicholaspai:
- APR does not change based on inputted share of pool, should always display
- Removed hard coded values in reward token object
Screen Shot 2020-09-09 at 10 10 19

### DIFF
--- a/features/yield/FarmingCalculator.tsx
+++ b/features/yield/FarmingCalculator.tsx
@@ -166,7 +166,7 @@ const FarmingCalculator = () => {
   // Update APR
   useEffect(() => {
     calculateApr();
-  }, [tokenAddress, poolLiquidity]);
+  }, [tokenAddress, poolLiquidity, rewardToken, rewardTokenPrices]);
 
   // Update pool ownership
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


features/yield/FarmingCalculator.tsx
```diff
@@ -31,356 +30,327 @@ const FarmingCalculator = () => {
...
   const [yUSDPrice, setyUSDPrice] = useState<string>("");
   const [yUSDAdded, setyUSDAdded] = useState<string>("");
   const [USDCAdded, setUSDCAdded] = useState<string>("");
   const [poolLiquidity, setPoolLiquidity] = useState<string>("");
-  const [weeklyUmaRewards, setWeeklyUmaRewards] = useState<number>(0);
-  const [weeklyRenRewards, setWeeklyRenRewards] = useState<number>(0);
 
   // Yield outputs:
   const [fracLiquidity, setFracLiquidity] = useState<string>("");
-  const [renYieldAmount, setRenYieldAmount] = useState<string>("");
```
nicholaspai, line 80: No need to hardcode these stats vars as "renPrice" or "umaPrice"

chrismaree:
This is awesome! Thanks for updating it and making it more general. Your pattern will help when we run other LM programs

chrismaree (approved):

PR #164, opened 2020-09-17
fix(redeem): MAX button on Redeem page sets incorrectly

nicholaspai:
This button is supposed to set the amount to redeem to either the user's balance or the total position debt, depending on which is larger.

However, the comparison is comparing two numbers that have had their precision potentially reduced, so we should be comparing the two numbers using the same precision.

### DIFF
--- a/containers/Token.ts
+++ b/containers/Token.ts
@@ -21,6 +21,7 @@ function useToken() {
   const [decimals, setDecimals] = useState<number | null>(null);
   const [allowance, setAllowance] = useState<number | "Infinity" | null>(null);
   const [balance, setBalance] = useState<number | null>(null);
+  const [balanceBN, setBalanceBN] = useState<BigNumber | null>(null);
 
   const getTokenInfo = async () => {
     if (contract) {
@@ -44,6 +45,7 @@ function useToken() {
       setName(name);
       setDecimals(decimals);
       setBalance(balance);
+      setBalanceBN(balanceRaw);
       setAllowance(allowance);
     }
   };
@@ -64,6 +66,7 @@ function useToken() {
     setName(null);
     setDecimals(null);
     setBalance(null);
+    setBalanceBN(null);
     setAllowance(null);
     getTokenInfo();
   }, [contract]);
@@ -91,6 +94,7 @@ function useToken() {
     symbol,
     decimals,
     balance,
+    balanceBN,
     allowance,
     setMaxAllowance,
   };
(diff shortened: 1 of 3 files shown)


features/manage-position/Redeem.tsx
```diff
@@ -120,7 +120,9 @@ const Redeem = () => {
     };
 
     const setTokensToRedeemToMax = () => {
-      if (tokenBalance > posTokens) {
+      // `tokenBalance` and `posTokens` might be incorrectly rounded,
+      // so we round them arbitrarily to the same precision to compare.
+      if (tokenBalance.toFixed(4) >= posTokens.toFixed(4)) {
```
daywiss, line 125: how come you dont use the toWei, parseFixed or BN to compare these? it seems like this could cause an edge case but is ok in most cases. Theres some other parts where balances are cast with Number as well, im surprised this doesnt cause more problems.

nicholaspai: Yeah problem is that `tokenBalance` is passed as a `number` from the Position container. This is non-ideal and all the containers should ideally be passing only BN's to the front-end components.

daywiss: i think there are other issues with doing comparisons between numbers by converting them to strings first. This seems to work in many cases but here is one that fails:
`9.99.toFixed(4) > 10.0.toFixed(4)  == true`

Its because its comparing the values lexicographically since toFixed returns a string

nicholaspai: hmm I agree. So instead we could add a `balanceBN` to the Token container and use the BN value of the token balance instead of the truncated number

nicholaspai: @daywiss I fixed this using the solution above, please check

containers/Token.ts
```diff
@@ -21,6 +21,7 @@ function useToken() {
   const [decimals, setDecimals] = useState<number | null>(null);
   const [allowance, setAllowance] = useState<number | "Infinity" | null>(null);
   const [balance, setBalance] = useState<number | null>(null);
+  const [balanceBN, setBalanceBN] = useState<BigNumber | null>(null);
```
nicholaspai, line 24: I pass the balance as a BN so that client React components can compare the raw BN instead of the truncated balance

daywiss (approved):
Ok i think this seems a lot better, thanks

mrice32 (approved):
LGTM!

PR #165, opened 2020-09-18
fix withdrawal countdown missing padding for minutes and seconds

akhanubis:
Tiny fix for countdown not adding a leading 0 for minutes and seconds < 10.

image


nicholaspai (approved):
This LGTM! Thanks for the contribution @akhanubis 

I tested out withdrawal requests on Kovan

chrismaree (approved):

PR #166, opened 2020-09-21
improve(farming-calc): Prepare for Oct roll, more scalable for future changes

nicholaspai:
This adds a data structure for storing information about rolling yield tokens which makes it easier to display yield calculations for future LM rewards during rolls. 

Pre roll or token is not part of a roll. This is the same as current:
Screen Shot 2020-09-21 at 13 42 20

Mid-roll, taking into account 2 pools:
Screen Shot 2020-09-21 at 13 42 35

Post-roll: current token has been rolled:
Screen Shot 2020-09-21 at 13 44 16


features/yield/FarmingCalculator.tsx
```diff
@@ -38,33 +54,34 @@ const FarmingCalculator = () => {
   );
 
   // Yield inputs:
+  const currentTokenPrice = getTokenPrice(tokenAddress);
```
nicholaspai, line 57: This was moved from lower down. FYI both `getTokenPrice` and `getPoolData` are wrappers around graphQL hooks and therefore cannot be called within other hooks like `useEffect`

features/yield/FarmingCalculator.tsx
```diff
@@ -158,10 +175,32 @@ const FarmingCalculator = () => {
...
+      setTokenAddress(address.toLowerCase());
+    }
+    if (Object.keys(WEEKLY_UMA_REWARDS).includes(tokenAddress)) {
+      setRewardToken(WEEKLY_UMA_REWARDS[tokenAddress]);
+    }
+
+    // Check if token should display roll information.
+    if (Object.keys(ROLL_REWARDS_SCHEDULE).includes(tokenAddress)) {
+      setRollFromTokenObj(ROLL_REWARDS_SCHEDULE[tokenAddress]);
+    } else {
```
nicholaspai, line 190: This is the main change here. `rollFromTokenObj` is set to non-null if there is mapping in the `ROLL_REWARDS_SCHEDULE` struct for the current token address (`tokenAddress`)

features/yield/FarmingCalculator.tsx
```diff
@@ -170,28 +209,23 @@ const FarmingCalculator = () => {
 
   // Update pool ownership
   useEffect(() => {
-    if (
-      sept20PoolData &&
-      rewardTokenPoolData &&
-      yUSDAdded === "" &&
-      USDCAdded === ""
-    ) {
-      const userShareFractionSept = sept20PoolData.userSharesFraction;
```
nicholaspai, line 179: These vars are renamed to be more generalized

features/yield/FarmingCalculator.tsx
mrice32: Just curious, why can't it be called within a `useEffect`?

features/yield/FarmingCalculator.tsx
```diff
@@ -240,17 +274,17 @@ const FarmingCalculator = () => {
...
-              yUSD-OCT20 are considered equally. What this means is that UMA
-              rewards are granted pro-rata as: (your USD contribution) / (total
-              SEP20 liquidity + total OCT20 liquidity). After August 28, 23:00
-              UTC, only LP contributions in the OCT20 pool will count towards
+              <strong>Update (09/23 @ 23:00 UTC):</strong> Before September
+              23rd, 23:00 UTC, LP contributions to either the OCT20 or the NOV20
+              are considered equally. What this means is that UMA rewards are
+              granted pro-rata as: (your USD contribution) / (total OCT20
+              liquidity + total NOV20 liquidity). After September 27, 23:00 UTC,
+              only LP contributions in the NOV20 pool will count towards
```
mrice32, line 287: One thing that's a little confusing to me is that we're encoding all these times in code everywhere else, but we're hardcoding them in this text. Is there a quick way we can describe the general process of the roll and then add a list or table of applicable tokens with their roll status. This may be annoying, though, so I wouldn't worry about it unless you can think of a quick way to encode this.

mrice32 (approved):
I like this a lot. Very nice

features/yield/FarmingCalculator.tsx
chrismaree: Another option could be to store these strings in the `ROLL_REWARDS_SCHEDULE` object and then pull them out in the sentences.

features/yield/FarmingCalculator.tsx
```diff
@@ -259,13 +293,15 @@ const FarmingCalculator = () => {
         <br></br>
         <Typography>
           <strong>Pools eligible for liquidity mining:</strong>{" "}
-          {isRolled ? "yUSD-OCT, uUSDrBTC-OCT" : "SEPT + OCT"}
+          {timeUntilRoll
+            ? isRolled
+              ? "NOV yield tokens"
```
chrismaree, line 298: I would also perhaps want these dates to be stored within `ROLL_REWARDS_SCHEDULE`

chrismaree (approved):
Looks great! just have one comment on re-using variables from the `ROLL_REWARDS_SCHEDULE

daywiss:
So the idea here is that you just add a new data structure at the top and everything else should just work, is that right?

features/yield/FarmingCalculator.tsx
nicholaspai: Calling hooks within hooks violates the rule of hooks (that was a mouthful): https://github.com/facebook/react/issues/16582#issuecomment-526807226

I had this same error

features/yield/FarmingCalculator.tsx
nicholaspai: I can just read them from the ROLL_REWARDS_SCHEDULE start and end dates. Will fix

features/yield/FarmingCalculator.tsx
nicholaspai: Will remove the hard codes

nicholaspai:
> So the idea here is that you just add a new data structure at the top and everything else should just work, is that right?

That's the dream, and will help future roll's

daywiss (approved):

features/yield/FarmingCalculator.tsx
nicholaspai: I now read the token symbol dynamically and display either the current pool or the current pool + roll pool

features/yield/FarmingCalculator.tsx
nicholaspai: I now read the roll date dynamically

pages/index.tsx
```diff
@@ -59,7 +59,7 @@ export default function Index() {
     options.push("yUSD Yield");
   }
 
-  if (weth && collAddress == weth.address) {
+  if (weth && collAddress?.toLowerCase() == weth.address.toLowerCase()) {
```
daywiss, line 62: I haven actually used this new feature yet. What is this doing exactly, only calling toLower if collAddress it not null?

nicholaspai: yep, otherwise the LHS of this equation will be `undefined`

PR #169, opened 2020-09-24
improve(performance) & fix(farming calc): Remove unneccessary queries and fix farming calc fix

### LINKED ISSUE #167: dApp should not request data for components that are not rendered
Currently the dApp is making a lot of unneccessary subgraph queries which is slowing down the entire dApp.

Additionally, queued up queries for prevously selected EMPs should be cancelled. Otherwise a queue of requests can build up which exacerbates the slowness

### LINKED ISSUE #168: Farming calculator not refreshing when switching from a non-LM token to a LM token
Non-LM token like the Dec contracts should not display the Farming calc
Screen Shot 2020-09-24 at 07 47 26

But this calc appears when switching from a LM token like the Oct contracts to the Dec contracts
Screen Shot 2020-09-24 at 07 48 09

nicholaspai:
- Farming calc page was not resetting `rewardToken = null` when the selected EMP changes, which falsely displayed the Farming calc for non eligible tokens
- CoinGeckoAPI was getting pinged constantly (> 100 a second at least), I added a check to see if we previously loaded token price data
- GraphQL queries were refreshing every few seconds in the background, I removed this polling functionality

Fixes #168 
Fixes #167

### DIFF
--- a/containers/Balancer.ts
+++ b/containers/Balancer.ts
@@ -302,7 +302,6 @@ const useBalancer = () => {
     usdPrice,
     shares,
     userShareFraction,
-    initializeTokenAddress,
     getPoolDataForToken,
     getTokenPrice,
   };
--- a/containers/Balancer.ts
+++ b/containers/Balancer.ts
@@ -63,10 +63,6 @@ const useBalancer = () => {
     null
   );
 
-  // Because apollo caches results of queries, we will poll/refresh this query periodically.
-  // We set the poll interval to a very slow 5 seconds for now since the position states
-  // are not expected to change much.
-  // Source: https://www.apollographql.com/docs/react/data/queries/#polling
   const {
     loading: tokenPriceLoading,
     error: tokenPriceError,
@@ -75,14 +71,12 @@ const useBalancer = () => {
     skip: !selectedTokenAddress,
     context: { clientName: "BALANCER" },
     variables: { tokenId: selectedTokenAddress },
-    pollInterval: 5000,
   });
   const { loading: poolLoading, error: poolError, data: poolData } = useQuery(
     POOL(JSON.stringify(poolTokenList)),
     {
       skip: !poolTokenList,
       context: { clientName: "BALANCER" },
-      pollInterval: 5000,
     }
   );
 
--- a/containers/Balancer.ts
+++ b/containers/Balancer.ts
@@ -6,6 +6,7 @@ import { TOKENS, POOL } from "../apollo/balancer/queries";
 
 import Connection from "./Connection";
 import Token from "./Token";
+import { YIELD_TOKENS } from "../utils/yieldTokenList";
 
 interface PoolState {
   exitsCount: number;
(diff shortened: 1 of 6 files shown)


containers/Balancer.ts
```diff
@@ -6,6 +6,7 @@ import { TOKENS, POOL } from "../apollo/balancer/queries";
 
 import Connection from "./Connection";
 import Token from "./Token";
+import { YIELD_TOKENS } from "../utils/yieldTokenList";
```
nicholaspai, line 9: I moved this YIELD_TOKENS struct outside the Balancer container because I didn't want to force components to load all of the balancer data just to check whether a token address was a yield token

containers/Balancer.ts
```diff
@@ -104,14 +71,12 @@ const useBalancer = () => {
     skip: !selectedTokenAddress,
     context: { clientName: "BALANCER" },
     variables: { tokenId: selectedTokenAddress },
-    pollInterval: 5000,
```
nicholaspai, line 107: This removes the polling, this might be nice in the future but I don't see many user complaints about non-refreshing data.

features/yield/FarmingCalculator.tsx
```diff
@@ -137,13 +137,21 @@ const FarmingCalculator = () => {
...
       }
       const usdRewards: rewardTokenMap = {};
-      const tokenPrices: rewardTokenMap = {};
+
+      // Fetch token prices if not set already.
+      let tokenPrices: rewardTokenMap = rewardTokenPrices;
       for (let rewardObj of rewardToken) {
-        const _tokenPrice = await rewardObj.getPrice();
-        tokenPrices[rewardObj.token] = _tokenPrice;
+        let _tokenPrice;
```
nicholaspai, line 144: This code was unintentionally calling the coinGeckoApi frequently to load new `_tokenPrices` into the `rewardTokenPrices` state var. 

Here, I
- set `tokenPrices` by default to the existing `rewardTokenPrices` struct
- if `tokenPrices[token]` already exists, no need to call the API again
- Otherwise, call the API and load it into the state

features/yield/FarmingCalculator.tsx
```diff
@@ -186,6 +194,8 @@ const FarmingCalculator = () => {
     }
     if (Object.keys(WEEKLY_UMA_REWARDS).includes(tokenAddress)) {
       setRewardToken(WEEKLY_UMA_REWARDS[tokenAddress]);
+    } else {
+      setRewardToken(null);
```
nicholaspai, line 198: This hides the farming calc if there is no WEEKLY_UMA_REWARDS info for the token address

pages/index.tsx
```diff
@@ -50,8 +52,12 @@ export default function Index() {
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
   const [selectedIndex, setSelectedIndex] = useState(0);
   const { address: collAddress } = Collateral.useContainer();
+  const { address: tokenAddress } = Token.useContainer();
   const { contract: weth } = WethContract.useContainer();
-  const { isYieldToken } = Balancer.useContainer();
```
nicholaspai, line 54: Removing the Balancer.useContainer() call here makes sure that the dApp is not making any Balancer subgraph queries at the `index.tsx` level

containers/Balancer.ts
chrismaree: does this mean the only way to update the table is to hard reload it?

chrismaree (approved):

containers/Balancer.ts
nicholaspai: I think so, I'm OK with this though as its a UI tradeoff for performance

PR #170, opened 2020-09-24
improve(farming calc): can add start and end dates for LM reward tokens

nicholaspai:
- Rewards that have not started yet or are expired now display more helpful information


features/yield/FarmingCalculator.tsx
```diff
@@ -306,46 +359,51 @@ const FarmingCalculator = () => {
...
-                    {`- 1 ${rewardObj.token} = $${rewardTokenPrices[
-                      rewardObj.token
-                    ].toLocaleString()}`}
-                  </span>
-                )
-              );
-            })}
-        </Typography>
-
+        {rewardToken.length > 0 ? (
```
nicholaspai, line 362: This moves the entire `Typography` block inside this conditional. If there are no reward tokens to display, simply display to user that rewards are not currently available for this token

chrismaree (approved):
LGTM! good generalization

PR #176, opened 2020-09-27
fix(farming-calc) Add timezone and refine roll token messaging

nicholaspai:
Responding to user feedback:

Screen Shot 2020-09-27 at 01 19 20


features/yield/FarmingCalculator.tsx
```diff
@@ -65,12 +65,9 @@ const FarmingCalculator = () => {
   );
   const rollToPool = getPoolDataForToken(rollToTokenAddress);
   const rollStartDate = rollFromTokenObj && rollFromTokenObj.rollStartDate;
-  const rollFromTokenName = isRollToToken
-    ? rollFromTokenObj && rollFromTokenObj.rollFromTokenName
-    : rollFromTokenObj && rollFromTokenObj.rollToTokenName;
-  const rollToTokenName = !isRollToToken
-    ? rollFromTokenObj && rollFromTokenObj.rollFromTokenName
-    : rollFromTokenObj && rollFromTokenObj.rollToTokenName;
+  const rollFromTokenName =
```
nicholaspai, line 68: @chrismaree what was the reasoning behind the `isRollToToken` logic? It seemed to be showing the wrong `rollToTokenName` when selecting the `rollFromToken` like the OCT tokens.

chrismaree: The logic was to ensure that we always had the right combination of `from` and `to` token names during the roll.

chrismaree (approved):
LGTM.

features/yield/FarmingCalculator.tsx
nicholaspai: This actually worked mid-roll

nicholaspai: I tested it out pre-roll, post-roll, mid-roll. I think it probably reverses if you don't supply the `rollToToken` heh

PR #178, opened 2020-09-27
improve(yield): Display temporary yield page redirecting to balancer pools as long as balancer subgraph is returning inaccurate information

nicholaspai:
Screen Shot 2020-09-27 at 21 28 44


mrice32 (approved):
LGTM!

PR #183, opened 2020-09-29
improve(farming-calc) revert temp calc page

nicholaspai:
Balancer subgraph appears correct, reverting temp page now


chrismaree (approved):
looks good to me! I verified both pools against both pools.balancer.exchange page and the balancer.exchange pages to verify the quoted prices are correct.

daywiss (approved):

PR #184, opened 2020-09-29
improve(expiry): Should check for `expiryPrice` on EMP contract in addition to DVM resolved price

nicholaspai:
It is possible that DVM has not resolved a price but EMP has an `expiryPrice()` (if the DVM has been upgraded since the EMP's expiry), and its possible that an EMP has no `expiryPrice()` but the DVM has resolved a price (right after the DVM has finished its reveal round but before anyone has attempted to `settle()` on the EMP)


chrismaree (approved):
looks good! I tested on the staging link as well and everything seems good to go

PR #186, opened 2020-09-30
feat&fix(liquidation-liveness) add liquidation liveness to General Info and fix liquidation table

### LINKED ISSUE #185: Incorrect liquidations table with diffrent liveness periods
A position was liquidated in the uUSDrBTC OCT contract, as can be seen in the screenshot below:
image

The Liquidation Status reads "Liquidation Succeeded". However, there is still a liquidation liveness timer remaining, as can be seen here: 

image

chrismaree:
This PR does two things:
1) The liquidation liveness is not shown on the "General Info" table. Normally, this is the same as withdrawal liveness but this is not strictly the case. This PR adds this information.

After this PR the dispute params looks as follows.
image

2) there was an error in the liquidations table, as documented in #185 where the `LiquidationActionDialog` incorrectly displayed the liquidation time remaining due to incorrect variable usage.

close #185

### DIFF
--- a/features/all-positions/LiquidationActionDialog.tsx
+++ b/features/all-positions/LiquidationActionDialog.tsx
@@ -133,7 +133,7 @@ const LiquidationActionDialog = (props: DialogProps) => {
 
     const liquidationTimeRemaining =
       Number(liquidatedPosition.liquidationTimestamp) +
-      Number(withdrawalLiveness) -
+      Number(liquidationLiveness) -
       Number(currentTime);
 
     const pendingLiquidationTimeString =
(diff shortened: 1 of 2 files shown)


features/all-positions/LiquidationActionDialog.tsx
```diff
@@ -133,7 +133,7 @@ const LiquidationActionDialog = (props: DialogProps) => {
 
     const liquidationTimeRemaining =
       Number(liquidatedPosition.liquidationTimestamp) +
-      Number(withdrawalLiveness) -
+      Number(liquidationLiveness) -
```
nicholaspai, line 136: Ah good catch, but why would this have been an error if withdrawal liveness and liquidation liveness are the same value?

nicholaspai (approved):
Great catch!

features/all-positions/LiquidationActionDialog.tsx
chrismaree: would not have been an issue if they were the same.

PR #188, opened 2020-10-13
fix(yield): changed the way APY is calculated, add jest for unit tests

### LINKED ISSUE #187: Investigate and correct emp-tools APY calculator
**Description**

Essentially the APY calculation in emp tools is not correct. Need to investigate and find right solution.

**Reproduction**
See apy calculator 

**Expected Behavior**

Should should correct APY calculation

May be able to use this formula:
> Following along for the  annualised number for the 10% case going from 0.9 to 1 as if you are buying yUSD at 0.9 and holding until settlement. These are the equations I think it should be:
> ((1+0.1)^(365/10)-1)*100 = 3142.149
> APY = ((1+(1-yUsdPrice))^(365/daysToSettlement)-1)*100

**Screenshots**

> No matter how much I sell yUSD for the max is 100%  APR
image

Additional Sources

This issue is discussed in slack:
https://risklabs.slack.com/archives/C01090AUD38/p1600191292001500

daywiss:
This PR adds compounding interest calculator and uses that to calculate APY. It also renames APR to APY.

It also adds several dependencies to allow for unit testing. This theoretically would allow testing of react components, but more work would be needed in order for that to work 100%.

see: https://www.omnicalculator.com/finance/apy
image
compare to:
image

image
compare to:
image

Closes #187

### DIFF
--- a/features/yield/YieldCalculator.tsx
+++ b/features/yield/YieldCalculator.tsx
@@ -4,6 +4,7 @@ import styled from "styled-components";
 import EmpState from "../../containers/EmpState";
 import Balancer from "../../containers/Balancer";
 import Token from "../../containers/Token";
+import { calcInterest, calcApy } from "../../utils/calculators";
 
 const FormInput = styled.div`
   margin-top: 20px;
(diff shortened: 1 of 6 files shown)


chrismaree:
high-level comment: the calculations you are using here (and the omnicalculator tool) are running at a compounding frequency of once per day. Does this make sense for this kind of production? should it be continuously compounding? or should there be _no_ compounding as there is no "re-investment" & as such should simply be the gain over the duration (i.e the compounding period is the term)?

daywiss:
> high-level comment: the calculations you are using here (and the omnicalculator tool) are running at a compounding frequency of once per day. Does this make sense for this kind of production? should it be continuously compounding? or should there be no compounding as there is no "re-investment" & as such should simply be the gain over the duration (i.e the compounding period is the term)?

Im going to add an APR calculator and try that instead, might make more sense

nicholaspai (commented):
Love that you created a test framework with potential to add more tests in the future. I also like the architecture you're using and think its very clean. I was a bit worried when I saw the line diff but happy that's all in the `.lock` file. To help me test, which formula are you using from here? https://www.calculatorsoup.com/calculators/financial/compound-interest-calculator.php

I also agree you should use continuous compounding if its practical to do so

daywiss:
> To help me test, which formula are you using from here? https://www.calculatorsoup.com/calculators/financial/compound-interest-calculator.php

Using the Rate (R) formula, this was to calculate the interest rate, which i then plug into the APY calculation. Pushing up an APR calculation that is simpler and might be a better representation.

nicholaspai (approved):
Looks great, so is it expected that APR for buyers and sellers would be different?

chrismaree (approved):
I think this is great! using the simple interest equation the way you have it set up here makes the most sense to me: _one_ compounding period over the full duration of the contract. easy to reason about, easy for users to verify, and fits within the other defi yield farming calculators definition of apy & apr.

Also really like your testing. future prs should add some of these kinds of tests to other utils. 

Lastly, having the compounding function `calcInterest` (even though it's not used) will be useful in the future :)

daywiss:
> Looks great, so is it expected that APR for buyers and sellers would be different?
Yes i think its to be expected. Will be slightly different

Issue #191, opened 2020-10-16
WalletConnect txs fail on frontend


rhlsthrm:
When I try to do anything I get this error: 
`The method eth_sendTransaction does not exist/is not available`

This could be due to outdated packages?

rhlsthrm:
I spoke to some of the WalletConnect team about this issue, they said it's most likely due to the Infura key not being valid.

chrismaree:
Thanks for finding this and reporting the bug! I've upgraded onboard.js which seemed to fix it in my testing. Please let me know if this has not resolved your issues :)

rhlsthrm:
Just tested it, it works! Thank you for the fix!

PR #192, opened 2020-10-19
fix(wallet-connector) upgrade onboard.js

chrismaree:
This PR is in response to issue #191 to fix connection issues with wallet connect. After upgrading the version of onboard.js wallet connect works again within the connector. 

I have tested this in Chrome, Safari, and Firefox and was able to generate and send transactions from all browsers.


nicholaspai (approved):
Nice simple fix

PR #199, opened 2020-12-18
Adding march yield dollars

smb2796:


nicholaspai (approved):
Looks good, you're going to want to add these contracts to this [file](https://github.com/UMAprotocol/emp-tools/blob/master/constants/yieldTokens.ts) in order to get the "Yield" page to render properly.

I checked out the vercel link and the Manage Position + All Positions tabs look good.

nicholaspai:
image (5)

The Yield tab now renders an empty table when the Balancer subgraph does not have a `tokenPrices` field for a specific pool. This is possible before the pool has had any trades. Previously, the dApp would crash when there was no `tokenPrices` field, now we catch it and show an empty table.

constants/rewards.ts
```diff
@@ -47,22 +47,38 @@ export const WEEKLY_UMA_REWARDS: { [key: string]: any[] } = {
       endDate: Date.UTC(2020, 9, 1, 0, 0, 0),
     },
   ], // uUSDrBTC-DEC
+  "0x90f802c7e8fb5d40b0de583e34c065a3bd2020d8": [
+    {
+      token: "UMA",
+      count: 15000,
```
nicholaspai, line 53: It'd be helpful to add a comment here about how this is an approximation based on expected dev mining allocation

constants/rewards.ts
```diff
@@ -47,22 +47,38 @@ export const WEEKLY_UMA_REWARDS: { [key: string]: any[] } = {
...
-    rollDate: Date.UTC(2020, 8, 27, 23, 0, 0, 0),
-  }, // yUSDETH-Oct20 --> uUSDwETH-DEC
-  "0x208d174775dc39fe18b1b374972f77ddec6c0f73": {
-    rollFromTokenName: "uUSDrBTC-OCT",
-    rollToTokenName: "uUSDrBTC-DEC",
-    rollToToken: "0xf06ddacf71e2992e2122a1a0168c6967afdf63ce",
-    rollStartDate: Date.UTC(2020, 8, 23, 23, 0, 0, 0),
-    rollDate: Date.UTC(2020, 8, 27, 23, 0, 0, 0),
-  }, // uUSDrBTC-OCT --> uUSDrBTC-DEC
+  "0xd16c79c8a39d44b2f3eb45d2019cd6a42b03e2a9": {
```
nicholaspai, line 70: So this is going to display that LM rewards are shared between YD and uUSD between 12/25-12/28 is that correct?

smb2796: Yes, I believe so. @daywiss can you confirm?

constants/rewards.ts
smb2796: Done

constants/rewards.ts
daywiss: I'm actually not sure, will need to check what i did when i updated estimate, but i have a feeling I did not include the roll calculation

daywiss: I'm not really familiar with how the roll logic works, but its still in there unchanged, so it should work the same as before I changed the LM calculations. Would need to run a test somehow to guarantee the correct behavior.

constants/rewards.ts
```diff
@@ -47,22 +47,38 @@ export const WEEKLY_UMA_REWARDS: { [key: string]: any[] } = {
...
+      token: "UMA",
+      count: 15000, // This is an approximation based on expected dev mining allocation
+      getPrice: getUmaPrice,
+      startDate: Date.UTC(2020, 12, 25, 23, 0, 0, 0),
+    },
+  ], // YD-ETH-MAR21
+  "0x002f0b1a71c5730cf2f4da1970a889207bdb6d0d": [
+    {
+      token: "UMA",
+      count: 10000, // This is an approximation based on expected dev mining allocation
```
daywiss, line 61: Since LM mining estimation takes into account dev mining, this value does not need to be hardcoded, but it doesnt harm anything leaving it in. The estimation does assume 50% of dev mining goes to LM

daywiss (approved):
I'm not well versed with adding tokens here but looks good.

PR #201, opened 2020-12-26
fix(farming-calculator): Fix how dev mining rewards are calculated mid-roll + other small fixes

nicholaspai:
- The month in `Date.UTC()` is supposed to be zero-indexed
- The farming calc hard-codes liquidity rewards to 1/2 of dev mining rewards, but mistakenly computes dev mining rewards for the current EMP address even when the roll is currently undergoing. In these situations, the dev mining rewards should be computed for the EMP from which the roll is occurring (i.e. the "old" EMP)
- Adds an end-date to DEC contracts so that they stop showing rewards post-roll.

Overall, this entire liquidity mining calculator clearly needs a makeover.

### DIFF
--- a/constants/rewards.ts
+++ b/constants/rewards.ts
@@ -30,6 +30,7 @@ export const WEEKLY_UMA_REWARDS: { [key: string]: any[] } = {
       count: 25000,
       getPrice: getUmaPrice,
       startDate: Date.UTC(2020, 8, 23, 23, 0, 0, 0),
+      endDate: Date.UTC(2020, 11, 28, 23, 0, 0, 0),
     },
   ], // uUSDwETH-DEC
   "0xf06ddacf71e2992e2122a1a0168c6967afdf63ce": [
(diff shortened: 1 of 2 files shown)


smb2796:
Thanks for taking care of this fix. Estimated rewards for WETH contracts look disproportionately small to me: BTC contracts display 7 mil in synths and 10,000 UMA per week, WETH contracts display 9 mil in synths and 7,000 UMA per week. Took a look around and couldn't figure out why this was happening. Any ideas?

nicholaspai:
The LMrewards are hard coded to 1/2 of the dev mining rewards. So I'll have to look into the dev mining Container

nicholaspai:
> Thanks for taking care of this fix. Estimated rewards for WETH contracts look disproportionately small to me: BTC contracts display 7 mil in synths and 10,000 UMA per week, WETH contracts display 9 mil in synths and 7,000 UMA per week. Took a look around and couldn't figure out why this was happening. Any ideas?

Also, the WETH contracts total closer to 15mil in synths between the DEC and MAR pools and BTC is closer to 8mil so this actually makes sense.

smb2796 (approved):
Looks good to me then!

PR #203, opened 2021-02-04
fix(calculators): update EMP valuation to use synthetic, and import active emp list

### LINKED ISSUE #202: "Estimated UMA LM Payout" on EMP tools should reflect a more accurate estimation to actual LM payouts
We've been getting several complaints about the APR estimator on EMP tools for liquidity mining. There are 2 aspects to the bad estimation:
1. We are not dynamically updating all active EMPs in dev mining
2. EMP valuation is currently using TVL rather than TVM, this is due to limitations on getting prices for synthetics.

Suggested solution:
1. We have added a protocol affilites file which is updated with active emps in dev mining. Pull that in to get the emp list.
2. We should also attempt to use TVM, pulling synthetic prices from coingecko and falling back to 1 USD when price is not available.  This is fragile but will work for now
3. Eventually get up dedicated APR calculator backend and switch to this.

daywiss:

Motivation #202 
This makes 2 major changes to EMP calculation:
- DevMining container now imports the active emplist from our github which @smb2796 and I maintain
- Previous EMP valuation was based off of total collateral locked. This changes it to synthetic minted, and attempts to get a price from coingecko, otherwise fallsback to 1 dollar, which most synthetics are. 

This produces much better results and are very close to this weeks payouts:
- yd eth estimates 7,886.682  vs [payout ](https://github.com/UMAprotocol/protocol/pull/2499)of 7080 
- yd btc estimates 8,681.799 vs [payout ](https://github.com/UMAprotocol/protocol/pull/2500)of 9067 

Closes #202

### DIFF
--- a/containers/DevMining.ts
+++ b/containers/DevMining.ts
@@ -9,26 +9,34 @@ import uma from "@studydefi/money-legos/uma";
 
 import Connection from "./Connection";
 
-export const defaultEmpWhitelist = [
-  "0xaBBee9fC7a882499162323EEB7BF6614193312e3",
-  "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56",
-  "0x3a93E863cb3adc5910E6cea4d51f132E8666654F",
-  "0xE4256C47a3b27a969F25de8BEf44eCA5F2552bD5",
-  "0x1c3f1A342c8D9591D9759220d114C685FD1cF6b8",
-];
-
 export const defaultTotalRewards = 50000;
+const empStatusUrl =
+  "https://raw.githubusercontent.com/UMAprotocol/protocol/master/packages/affiliates/payouts/devmining-status.json";
 
-const useDevMiningCalculator = ({
-  totalRewards = defaultTotalRewards,
-  empWhitelist = defaultEmpWhitelist,
-} = {}) => {
+const useDevMiningCalculator = () => {
   const { provider } = Connection.useContainer();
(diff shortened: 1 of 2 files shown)


containers/DevMining.ts
```diff
@@ -9,26 +9,34 @@ import uma from "@studydefi/money-legos/uma";
...
   const { provider } = Connection.useContainer();
   const [devMiningRewards, setRewards] = useState<Map<string, string> | null>();
   const [devMiningCalculator, setCalculator] = useState<any | null>();
+  const [empWhitelist, setEmpWhitelist] = useState<string[]>();
+  const [totalRewards, setTotalRewards] = useState<number>(defaultTotalRewards);
+
+  // pull latest whitelist
+  useEffect(() => {
+    fetch(
+      "https://raw.githubusercontent.com/UMAprotocol/protocol/master/packages/affiliates/payouts/devmining-status.json"
```
nicholaspai, line 24: When would you expect this URL to change? If any, just leave a comment otherwise can ignore this comment.

nicholaspai (approved):
Thanks for the quick fix, this should be more robust

containers/DevMining.ts
daywiss: it is not meant to change. it was set up specifically for this purpose and other projects are also relying on this for their APR calcs

mrice32 (approved):
LGTM!

PR #206, opened 2021-02-12
fix(emp-tools): change cors proxy and cleanup other requests

### LINKED ISSUE #205: cors errors reaching bitstamp price api
image
It looks like the cors proxy we were using expired: `https://cors-anywhere.herokuapp.com/https://www.bitstamp.net/api/v2/ticker/btcusd`

@mrice32 was able to spin up our own on gcp. need to update emp-tools to use this. one small caveat is it wont accept the protocol part of the url, so https will need to be stripped before proxying.

daywiss:

motivation #205 

Simply adds new cors proxy wrapper to bitstamp url. Adds as helper function in case we need to do this for other urls as well. 
Also updates coingecko calls to remove uncessessary header info, did not run into issues after this change, but verify if you can. 

fixes #205

### DIFF
--- a/utils/getCoinGeckoTokenPrice.ts
+++ b/utils/getCoinGeckoTokenPrice.ts
@@ -3,17 +3,15 @@
 
 // so this can be tested outside browser
 import fetch from "isomorphic-fetch";
+const fetchOptions = {
+  // currently no options needed, previously we were setting content type and accept.
+};
 
 export async function getUmaPrice() {
   const query =
     "https://api.coingecko.com/api/v3/simple/price?ids=uma&vs_currencies=usd";
 
-  const response = await fetch(query, {
-    headers: {
-      Accept: "application/json",
-      "Content-Type": "application/json",
-    },
-  });
+  const response = await fetch(query, fetchOptions);
 
   let priceResponse = await response.json();
   return priceResponse.uma.usd;
@@ -23,12 +21,7 @@ export async function getRenPrice() {
   const query =
     "https://api.coingecko.com/api/v3/simple/price?ids=republic-protocol&vs_currencies=usd";
 
-  const response = await fetch(query, {
-    headers: {
-      Accept: "application/json",
-      "Content-Type": "application/json",
-    },
-  });
+  const response = await fetch(query, fetchOptions);
 
   let priceResponse = await response.json();
   return priceResponse["republic-protocol"].usd;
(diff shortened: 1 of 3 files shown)


utils/getCoinGeckoTokenPrice.ts
```diff
@@ -3,17 +3,15 @@
...
 import fetch from "isomorphic-fetch";
+const fetchOptions = {
+  // currently no options needed, previously we were setting content type and accept.
+};
 
 export async function getUmaPrice() {
   const query =
     "https://api.coingecko.com/api/v3/simple/price?ids=uma&vs_currencies=usd";
 
-  const response = await fetch(query, {
```
mrice32, line 11: Why remove these? Did they not work with these options? Or is this just not required anymore?

utils/getOffchainPrice.ts
```diff
@@ -30,6 +30,19 @@ function _getBitstampPriceFromJSON(jsonData: any) {
...
 
+// This is needed because our new cors proxy wont forward http or https. We really only use https.
+function stripProtocol(url: string) {
+  return url.replace("https://", "");
+}
+
+// Small helper to wrap a url in cors proxy.
+function proxyUrl(
+  url: string,
+  proxy = "https://us-central1-uma-protocol.cloudfunctions.net/cors-proxy"
```
mrice32, line 41: Long-term, we should probably have this proxy address passed in as a build env variable.

mrice32 (approved):
LGTM!

utils/getCoinGeckoTokenPrice.ts
daywiss: While testing out this cors stuff, i removed the headers and did not notice any detrimental effects, It was working fine with them, but it just seemed unnecessary so I left them out.

utils/getOffchainPrice.ts
daywiss: not sure if this project will reference a .env, didnt try. If it does though it should be easy to add

nicholaspai (approved):
LGTM

utils/getCoinGeckoTokenPrice.ts
```diff
@@ -3,17 +3,18 @@
 
 // so this can be tested outside browser
 import fetch from "isomorphic-fetch";
+const fetchOptions = {
+  headers: {
```
daywiss, line 7: i added the header options back in. rather not introduce potential bug.

utils/getOffchainPrice.ts
```diff
@@ -30,6 +30,22 @@ function _getBitstampPriceFromJSON(jsonData: any) {
...
+function stripProtocol(url: string) {
+  return url.replace("https://", "");
+}
+
+// Small helper to wrap a url in cors proxy.
+export function proxyUrl(
+  url: string,
+  // This env can be set with a .env file. It has to be prefixed with NEXT_PUBLIC_ to be available in browser
+  // This has a default set for now until we can get prod env var in.
+  proxy = process.env.NEXT_PUBLIC_CORS_PROXY_URL ||
```
daywiss, line 43: i tested this. and it does inject env in if set. left in hardcoded string so it works in production

utils/getOffchainPrice.ts
daywiss: added in optional env var. see comment

utils/getCoinGeckoTokenPrice.ts
daywiss: added this back in just in case

PR #213, opened 2021-03-19
 feat: add perpetual info boxes, clean up non working components

### LINKED ISSUE #204: Rework EMP tools to work with Perp
Created by David Adams via monday.com integration. 🎉

daywiss:
# Motivation
#204 

# Summary
- Add perp specific boxes for fair value, funding rate and market price
- adds helper utilities for uniswap gql calls
- clears out perp general info component and totals due to incompatibility with emp state

This needs a lot more work to be production ready, but this is for future work.
- custom containers for getting perp specific state for the general info component
- fixing existing containers to not fetch data based on emp abis 
- position managment tab

# Screenshots
image

### DIFF
--- a/constants/perpetuals.ts
+++ b/constants/perpetuals.ts
@@ -1,0 +1,26 @@
+import assert from "assert";
+// hard coded perpetual info, until theres a better way to get this info
+export const perpetuals = [
+  {
+    // this is a perp kovan contract
+    name: "New Perpetual Contract Test",
+    symbol: "NEW-PERP-TEST",
+    // this is hardcoded to a mainnet uniswap eth usdc pair for now. But in the future we should be able
+    // to derive the market or markets where the synthetic is being traded...
+    market: {
+      type: "uniswap",
+      id: "0xbb2b8038a1640196fbe3e38816f3e67cba72d940",
+      token0: "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599",
+      token1: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
+    },
+  },
+];
+
+export function findInfoByName(name: string | undefined) {
+  assert(name, "requires perpetual name");
+  const result = perpetuals.find(
+    (perp) => perp.name.toLowerCase() == name.toLowerCase()
+  );
+  assert(result, `Unable to find perp info by name: ${name}`);
+  return result;
+}
--- a/constants/perpetuals.ts
+++ b/constants/perpetuals.ts
@@ -5,6 +5,7 @@ export const perpetuals = [
     // this is a perp kovan contract
     name: "New Perpetual Contract Test",
     symbol: "NEW-PERP-TEST",
+    address: "0x24d15f2607ee56dF752375a63e646cbF8E652aF3",
     // this is hardcoded to a mainnet uniswap eth usdc pair for now. But in the future we should be able
     // to derive the market or markets where the synthetic is being traded...
     market: {
@@ -22,5 +23,14 @@ export function findInfoByName(name: string | undefined) {
     (perp) => perp.name.toLowerCase() == name.toLowerCase()
   );
   assert(result, `Unable to find perp info by name: ${name}`);
+  return result;
+}
+
+export function findInfoByAddress(address: string | undefined) {
+  assert(address, "requires perpetual address");
+  const result = perpetuals.find(
+    (perp) => perp.address.toLowerCase() == address.toLowerCase()
+  );
+  assert(result, `Unable to find perp info by address: ${address}`);
   return result;
 }
--- a/containers/PerpetualState.ts
+++ b/containers/PerpetualState.ts
@@ -1,0 +1,59 @@
+import { createContainer } from "unstated-next";
+import { useState, useEffect } from "react";
+import { BigNumber, Bytes, ethers } from "ethers";
+
+import Connection from "./Connection";
+import SelectedContract from "./SelectedContract";
+import EmpContract from "./EmpContract";
+import EmpState from "./EmpState";
+import { getAbi } from "../utils/getAbi";
+
+type Provider = ethers.providers.Provider | ethers.Signer;
+
+export async function getPerpetualState(address: string, provider: Provider) {
+  const abi = getAbi("perpetual", "2");
+  const contract = new ethers.Contract(address, abi, provider);
+  return {
+    fundingRate: await contract.fundingRate(),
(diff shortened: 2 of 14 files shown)


containers/EmpState.ts
```diff
@@ -50,83 +50,84 @@ const initState = {
   finderAddress: null,
   expiryPrice: null,
 };
+// get state from EMP
+export const queryState = async (emp: any) => {
```
daywiss, line 54: this was a pure function which I exported in case it could be useful somewhere else

constants/perpetuals.ts
```diff
@@ -0,0 +1,26 @@
+import assert from "assert";
+// hard coded perpetual info, until theres a better way to get this info
+export const perpetuals = [
+  {
+    // this is a perp kovan contract
+    name: "New Perpetual Contract Test",
+    symbol: "NEW-PERP-TEST",
+    // this is hardcoded to a mainnet uniswap eth usdc pair for now. But in the future we should be able
+    // to derive the market or markets where the synthetic is being traded...
```
nicholaspai, line 9: This will come in handy for the future to derive the market for a given pair: https://uniswap.org/docs/v2/SDK/pair/

constants/perpetuals.ts
```diff
@@ -0,0 +1,26 @@
...
+      id: "0xbb2b8038a1640196fbe3e38816f3e67cba72d940",
+      token0: "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599",
+      token1: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
+    },
+  },
+];
+
+export function findInfoByName(name: string | undefined) {
+  assert(name, "requires perpetual name");
+  const result = perpetuals.find(
```
nicholaspai, line 21: Maybe you should `findInfoByPerpAddress` instead of name?

containers/EmpState.ts
nicholaspai: Won't `expiryPrice` fail on a Perp? Or is the assumption that `queryState` would never be called on a Perp

containers/PerpetualState.ts
```diff
@@ -0,0 +1,59 @@
...
+
+type dataType = {
+  fundingRate: {
+    rate: BigNumber;
+  };
+  priceIdentifier: string;
+};
+const useContractState = () => {
+  const { signer } = Connection.useContainer();
+  const abi = getAbi("perpetual", "2");
```
nicholaspai, line 30: What's 2?

containers/PerpetualState.ts
```diff
@@ -0,0 +1,59 @@
...
+  const { address, isValid } = SelectedContract.useContainer();
+  const [state, setState] = useState<{
+    data: dataType | null;
+    error: Error | null;
+  }>({ data: null, error: null });
+  const [loading, setLoading] = useState<boolean>(true);
+  useEffect(() => {
+    if (!isValid) return;
```
nicholaspai, line 38:
```suggestion
    if (!isValid || !address || !signer) return;
```

containers/Uniswap.ts
```diff
@@ -1,20 +1,48 @@
...
     loading,
-    error,
     data,
+    error,
   };
-};
+}
 
-export default useUniswap;
+export const UniswapGetPair = createContainer(getPair);
```
nicholaspai, line 48: This is a cool pattern and reduces the need to type the main export's variable twice

features/perpetual/contract-state/GeneralInfo.tsx
```diff
@@ -32,160 +31,128 @@ const Status = styled(Typography)`
...
+      </Status>
+      <Status>
+        <Label>Minimum sponsor tokens: </Label>
+        {minSponsorTokensSymbol}
+      </Status>
+    </Box>
+  );
+}
+
+// TODO: price feed will need to be updated to use uniswap and also be able to calculate funding rate
```
nicholaspai, line 142: I wonder if we can export some of the pricefeed logic from @uma/financial-templates-lib, otherwise we're just gonna end up re writing everything.

utils/calculators.ts
```diff
@@ -165,3 +167,12 @@ export function DevMiningCalculator({
     },
   };
 }
+
+export function calculateFairValue(
+  fundingRate: BigNumber,
+  marketPrice: string
+) {
+  // convert market price to wei from a numerical string in eth
+  const marketPriceWei = parseEther(parseFloat(marketPrice).toFixed(18));
```
nicholaspai, line 176: Is it possible for `marketPrice` or `marketPriceWei.mul(fundingRate)` to not be in 18 precision?

containers/EmpState.ts
daywiss: yea this currently is not called with ther perp because there are incompatibilities. This is why the general info section in the screenshot is currently empty.

constants/perpetuals.ts
daywiss: yea i didnt know what else to do here, but ive included a subgraph call to get amm information based on the token pair or the id of the amm.

constants/perpetuals.ts
daywiss: thats a good idea

containers/PerpetualState.ts
daywiss: its a version number. I've set up a getAbi function which has some short names for version matching as well as contract name matching. you can see [here](https://github.com/UMAprotocol/emp-tools/blob/047163a3d2bd1f8dc674a4a8eb790c4dccd3cb28/utils/getAbi.ts#L29)

features/perpetual/contract-state/GeneralInfo.tsx
daywiss: that is the dream but i dont think we are quite there yet

utils/calculators.ts
daywiss: currently the marketprice is pulled from unis subgraph and it looks like this is always going to be in decimal form. it may be the case in the future though this assumption breaks. I will add a comment for it

constants/perpetuals.ts
```diff
@@ -0,0 +1,36 @@
...
+    (perp) => perp.name.toLowerCase() == name.toLowerCase()
+  );
+  assert(result, `Unable to find perp info by name: ${name}`);
+  return result;
+}
+
+export function findInfoByAddress(address: string | undefined) {
```
daywiss, line 29: ive updated lookups to use this instead of by name

mrice32 (approved):
Looks good!

PR #215, opened 2021-03-25
feat: get emp v2 contract state working

### LINKED ISSUE #214: Make sure EMP tools works with the new EMP
Created by David Adams via monday.com integration. 🎉

daywiss:

# motivation
#214

# summary
This allows emp contracts to work with different interfaces, mainly renaming of Pct functions to Percentage.  This switches the EmpState container to use the new and more general ContractState container under the hood. This leverages a new 'getState' function bound to each contract abi to allow both getting and mapping contract state into a form compatible with the views.

Another PR will be needed in order to fix decimalization between synthetic and collateral

### DIFF
--- a/utils/getAbi.ts
+++ b/utils/getAbi.ts
@@ -18,6 +18,7 @@ type ContractType = {
   abi: any[];
   getState?: (instance: Contract) => Promise<any>;
 };
+
 export const Contracts: ContractType[] = [
   {
     // we could include semver to do compatibilty checks for version strings, but it would add to bundle and this is good enough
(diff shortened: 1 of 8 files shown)


utils/getAbi.ts
```diff
@@ -1,38 +1,95 @@
...
     abi: emp2.abi,
+    async getState(instance: Contract) {
+      const state = {
+        expirationTimestamp: (await instance.expirationTimestamp()) as BigNumber,
+        collateralCurrency: (await instance.collateralCurrency()) as string, // address
+        priceIdentifier: (await instance.priceIdentifier()) as Bytes,
+        tokenCurrency: (await instance.tokenCurrency()) as string, // address
+        collateralRequirement: (await instance.collateralRequirement()) as BigNumber,
+        disputeBondPct: (await instance.disputeBondPercentage()) as BigNumber,
+        disputerDisputeRewardPct: (await instance.disputerDisputeRewardPercentage()) as BigNumber,
```
nicholaspai, line 66: I wonder if you don't need to repeat the same calls for methods that are common across versions, and then just make the different calls per version:

const state = {
...commonState,
sponsorDisputeRewardPct: (await instance.sponsorDisputeRewardPercentage()) as BigNumber,
...
}

nicholaspai (approved):
Well refactored, `getAbi` is a good place to add newer versions going forward. I've tested the dApp on mainnet and modified some test positions on the old EMP and all looks good

utils/getAbi.ts
daywiss: yea that is a really good suggestion, there really is only a handful of functions

PR #216, opened 2021-03-25
Update emp-tools to work with the new EMP

mrice32:
Small changes to ensure decimals are considered in all calculations.


daywiss (approved):
looks great, thanks @mrice32

PR #220, opened 2021-03-26
fix(rewards): Fix YD-BTC APR

smb2796:
This pr fixes the roll logic in `rewards.ts` to calculate YD-BTC-JUN's APY correctly.


daywiss (approved):
Thank you!

Issue #225, opened 2021-03-31
The Manage Postion tab throws "invalid BigNumber string" error for expired contracts


smb2796:
If a contract is expired, the `Manage Position` tab cannot be selected for that contract. The tab still works for contracts that are not expired. This is for both contracts that are currently still waiting for a price to be returned, or have already settled. Aka it applies to YD-BTC-MAR21 but also applies to uUSDwETH-DEC20.

To reproduce, connect your wallet, select any contract that has expired and then select `manage position`.

PR #226, opened 2021-03-31
fix: fix expiryPrice request in emp state causing app crash for expir…

daywiss:
…ed contracts

# motivation
#225 

# summary
This was caused by calling expiryPrice on the contract instance without actually invoking the function. small typo. various other small fixes found along the way to find the bug.

### DIFF
--- a/features/manage-position/ManagePosition.tsx
+++ b/features/manage-position/ManagePosition.tsx
@@ -21,17 +21,19 @@ const Manager = () => {
   const { empState } = EmpState.useContainer();
   const { isExpired } = empState;
   const [method, setMethod] = useState<Method>(DEFAULT_METHOD);
+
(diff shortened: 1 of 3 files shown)


features/manage-position/ManagePosition.tsx
```diff
@@ -21,17 +21,19 @@ const Manager = () => {
   const { empState } = EmpState.useContainer();
   const { isExpired } = empState;
   const [method, setMethod] = useState<Method>(DEFAULT_METHOD);
+
   const handleChange = (e: React.ChangeEvent<{ value: unknown }>) =>
     setMethod(e.target.value as Method);
 
+  useEffect(() => {
```
daywiss, line 28: useEffects should not be conditional, pulled out of if statement

smb2796 (approved):
Awesome find! LGTM

PR #235, opened 2021-05-28
fix: fix liquidator/disputer addresses to fix modal issue

### LINKED ISSUE #233: Liquidation modal is not accessible
Trying to access the liquidation modal throws a `TypeError: Cannot read property 'substr' of undefined` error and redirects back to the homepage.

To reproduce:

1. Select a contract with previous liquidations, such as YD-ETH-JUN21.
2. Navigate to `All Positions` -> `Show Liquidations`
3. Click on the kebab menu on one of the liquidations

image

4. This throws this error: `Error: Unable to calculate emp value, no token price or collateral price` and redirects back to the homepage.

Note: this happens with both v1.2.2 and 2.0 EMPs

daywiss:
fixes #233 
clubhouse: https://app.clubhouse.io/uma-project/story/970/disable-or-hide-the-ability-to-dispute-a-liquidation-from-emp-tools
**summary**
looks like issue was caused by minor change in subgraph api. liquidation and dispute addresses now are on `liquidator` and `disputer` rather than before `liquidator.address` and `disputer.address`

### DIFF
--- a/containers/EmpSponsors.ts
+++ b/containers/EmpSponsors.ts
@@ -43,8 +43,8 @@ interface LiquidationQuery {
   id: string;
   sponsor: { id: string };
   liquidationId: string;
-  liquidator: { address: string };
-  disputer: { address: string };
+  liquidator: string;
+  disputer: string;
   tokensLiquidated: string;
   lockedCollateral: string;
   liquidatedCollateral: string;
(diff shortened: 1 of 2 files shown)


smb2796 (approved):
Good find! LGTM

PR #237, opened 2021-06-08
Add DEXTF Kovan and Mainnet Yield Dollars 

smb2796:


utils/getOffchainPrice.ts
```diff
@@ -32,6 +32,10 @@ function _getBitstampPriceFromJSON(jsonData: any) {
   return Number(jsonData.last);
 }
 
+function _getCoingeckoPriceFromJSON(jsonData: any) {
+  return Number(jsonData.dextf.usd);
```
pemulis, line 36: Could this be generalized instead of being only for dextf?

smb2796: Yeah absolutely. No other contract uses CG here though, so didn't feel necessary to figure out. Thoughts? Think it needs to be generalized?

pemulis: I think it could be named `_getDextfPriceFromJSON` for now and then generalized to `_getCoingeckoPriceFromJSON` once we have multiple contracts that need CoinGecko prices?

pemulis: Either that, or add a `todo` comment that this should be generalized once more contracts are using CoinGecko.

smb2796: Added a todo comment. Thanks for bringing this up. I'll revisit later.

pemulis (approved):

